PR #5, opened 2017-11-29
WIP: Document converters

nokome:
Initial (re)implementations of Document converters


nokome:
@hamishmack I'll merge this now. This skips some of your tests because it uses pandoc 2.04 and not stencila/pandoc - once we have builds for all platforms there we can reinstate them.

PR #6, opened 2017-11-30
WIP: DocumentJupyterConverter

nokome:
The `DocumentJupyterConverter` class provides conversion between Jupyter Notebooks .ipynb JSON format and Stencila Documents

See previous iteration of this in https://github.com/stencila/stencila/pull/292

@hamishmack : this will need to call `DocumentMarkdownConverter` for doing the conversion of Markdown cells to/from JATS so will wait on that before progressing this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
   "dependencies": {
     "doctrine": "^2.0.0",
     "js-beautify": "^1.7.4",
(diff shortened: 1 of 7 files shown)


nokome:
Closing as implemented in https://github.com/stencila/convert/blob/master/src/document/DocumentJupyterConverter.js

PR #9, opened 2017-12-01
Remove bundling

nokome:
For simplicity make Node.js the only target for now.


nokome:
@hamishmack : I've reduced dependencies and a whole lot of indirection by make this a Node style project - so no transpilation or bundling. It means that we will need to change code to use `require` and `module.exports`. If your OK with this, could you merge this PR.

nokome:
Mmmm, after reading [this](http://2ality.com/2017/09/native-esm-node.html) I'm having second thoughts about converting to CJS modules - we might not need to do that.

PR #39, opened 2019-04-28
refactor: allow for supplying raw content

nokome:
This changes the API of several of the main functions:

- where appropriate change parameter `filePath` to `content`, which can be a string of raw content or a file path

- change parameter `mediaType` to `format` in places where either can be used and improve how the parameter is handled in `match`

This refactoring is mainly in preparation for adding compilers that act on plain text string, rather than files
- e.g. a compiler that detects a DOI and parses it into a `Article` by fetching meta data
- e.g. a compiler that [parses a string into a `Person`](https://github.com/stencila/schema/blob/fd19b920b48724731d5b1c7a7dd59b952f6d9caa/schema/organization/Person.ts#L15)
- moving the [implementation of the schema `parser` keyword](https://github.com/stencila/schema/blob/master/CONTRIBUTING.md#the-parser-keyword) into this repository (and moving all code out of the `schema` repo).

At the CLI it allows content to be provided as an argument, instead of in files or stdin e.g.

```bash
$ npx ts-node src/cli.ts "1,2,3" --from csv --to yaml
Converting from "1,2,3" to "yaml"

type: Datatable
name: Sheet1
columns:
  - type: DatatableColumn
    name: A
    values:
      - 1
  - type: DatatableColumn
    name: B
    values:
      - 2
  - type: DatatableColumn
    name: C
    values:
      - 3
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -101,20 +101,24 @@ interface Compiler {
  * string of raw content (e.g. `This is some Markdown`) or as a file system
  * path. This function, assesses whether a file is a file system path or not.
  *
- * If the string starts with `/`, `./` or `../` then it is assumed to be a path
- * but the presence of the path is not checked. For all other strings, this function
- * returns true if the file exists.
+ * If the string starts with `/`, `./`, `../` (or Windows compatible backslash
+ * equivalents as well as drive letter prefixed strings e.g. `C:\`)
+ * then it is assumed to be a path but the presence of the path is not checked.
+ * The reason for not checking presence
+ * here is because "if the content looks like a path then the user probably meant
+ * it to be a path". That is, if a user tries to convert the string "./file.md" then,
+ * if that file doesn't exist, she probably wants the Markdown converter to give me an error
+ * message. She probably doesn't want the `match` function to try and find some other converter
+ * that acts on a plain strings.
+ *
+ * For all other strings, this function does check for presence, returning `true`
+ * if the file exists.
  *
  * @param content The string to assess.
  */
 export function isPath(content: string): boolean {
-  if (
-    content.startsWith('/') ||
-    content.startsWith('./') ||
-    content.startsWith('../')
-  ) {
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -92,33 +95,66 @@ interface Compiler {
...
+ * but the presence of the path is not checked. For all other strings, this function
+ * returns true if the file exists.
+ *
+ * @param content The string to assess.
+ */
+export function isPath(content: string): boolean {
+  if (
+    content.startsWith('/') ||
+    content.startsWith('./') ||
+    content.startsWith('../')
```
alex-ketch: Could you share the reason for not checking the presence of the provided path please?
Reason being that these cases don't seem to take into account Windows file system paths. if we don't mind actually accessing the file system we could use [`fs.lstat`](http://nodejs.org/api/fs.html#fs_fs_lstat_path_options_callback), otherwise we might want to add `\` as an another separator.

src/index.ts
```diff
@@ -55,7 +56,7 @@ interface Compiler {
   /**
    * Any array of file names to use to match the compiler.
    * This can be useful for differentiating between
-   * flavours of formats like JSON e.g. `datapackage.json` versus any old `.json` file.
+   * "flavors" of formats e.g. `datapackage.json` versus any old `.json` file.
```
alex-ketch: 🇺🇸? 😋

src/index.ts
nokome: The reason for not checking presence here is because "if the content looks like a path then the user probably meant it to be a path". That is, if I try to convert the string "./file.md" then, if that file doesn't exist, I want the Markdown converter to give me an error message. I don't want the `match` function to try and find some other converter that acts on a plain strings. We use `fs.existsSync` in this function just to determine if a string that may or may not be a file, is actually a file. Does that make sense?

So for now I think I'd prefer to add \ as an another separator. Maybe make it a regex?

alex-ketch: Got it. Both the explanation and proposed solution makes sense. Thanks

nokome: Changes made accordingly. Thanks!

PR #40, opened 2019-04-29
Add bibliographic converters

nokome:
This PR adds several compilers related primarily to parsing citations and other bibliographic information. This will eventually allow for users to include reference lists in documents. 

It will be WIP while schemas for nodes created by these compilers e.g. `CreativeWork`, `Article` are developed over in https://github.com/stencila/schema.

The following new compilers target `Person` nodes (theses are relatively complete):

- `person`- parses and unparses a string representing a person (based on implementation in `schema` repo)
- `orcid` - parses an ORCID by fetching meta-data

The following new compilers target `CreativeWork` nodes:

- `csl` - parses CSL JSON
- `bib`- parses BibTeX
- `doi`- parses a DOI
- `crossref-query` - parses a biliographic search string

These are all based on the `csl` compiler which transforms CSL JSON into a `CreativeWork`. A lot more work is needed there. The `crossref-query` is experimental.

For more on using JSON-LD for bibliographies see:

- http://blog.schema.org/2014/09/schemaorg-support-for-bibliographic_2.html
- https://github.com/rdmpage/bibliographic-metadata-json#json-ld


src/codecs/csl/index.ts
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * Encode a `Node` to CSL-JSON or other bibliographic format
+ *
+ * See https://citation.js.org/api/tutorial-output_options.html
+ * for formats and other options which could be used.
+ */
+export const encode: Encode = async (
+  node: stencila.Node,
+  options: EncodeOptions = {}
+): Promise<vfile.VFile> => {
```
nokome: In the top level `encode` function we check the node type using type guards and either wrap it as needed e.g. `InlineContent` into a `CreativeWork` or log an `errorNodeType`

src/codecs/csl/index.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  return vfile.load(content)
+}
+
+/**
+ * Decode `Csl.Data` to a `CreativeWork` or derived type.
+ */
+export async function decodeCsl(
+  csl: Csl.Data
+): Promise<stencila.CreativeWork | stencila.Article> {
```
nokome: Decode function for the root node type which calls decode functions for child node types, in this case `decodeAuthor` and `decodeDate`

src/codecs/csl/index.ts
```diff
@@ -0,0 +1,228 @@
...
+    'container-title': containerTitle,
+    volume,
+    issue,
+    page,
+
+    // @ts-ignore // Citation.js uses this hidden property
+    _graph,
+
+    ...lost
+  } = csl
```
nokome: Use object destructing to concisely (a) set default values (instead of `const author = csl.author !== undefined ? csl.author : []`), and (b) get `lost` which is an object with all the properties that we don't explicitly handle below.

src/codecs/csl/index.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  if (type === 'article-journal') {
+    warnLossIfAny('csl', 'decode', csl, lost)
```
nokome: Call `warnLossIfAny` to emit a `log.warn` if there is anything in `lost`

src/codecs/csl/index.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    return stencila.article(authors, title, {
+      id,
+      datePublished,
+      isPartOf
+    })
+  } else {
+    warnLoss('csl', 'decode', `Unhandled citation type ${csl.type}`)
+    return stencila.creativeWork()
```
nokome: Rather than throwing an error. Just return an empty node.

src/codecs/csl/index.ts
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * Decode a `Csl.Person` as a `Person`
+ *
+ * If `family` is not defined but `literal` is then the
+ * `person` codec is used to decode the literal string.
+ *
+ * CSL-JSON's `non-dropping-particle` and `dropping-particle`
+ * are not currently supported in `Person`.
+ */
+const decodeAuthor = async (author: Csl.Person): Promise<stencila.Person> => {
```
nokome: Brief tl;dr type doc comments for each function - with more details if necessary. Don't worry about `@param` etc since TypeDoc does that for use / its self documenting having types.

src/codecs/doi/doi.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // The `foo` in these example makes them not DOIs
+  expect(await sniff('foo')).toBe(false)
+  expect(await sniff('doi: foo')).toBe(false)
+  expect(await sniff('doi: 10.1001/this/is/a/doi foo')).toBe(false)
+  expect(await sniff('http://foo.org/10.5334/jors.182')).toBe(false)
+})
+
+test('decode', async () => {
+  const { completeRecording } = await record('doi-decode')
```
nokome: Using `nock-record` to record HTTP responses in fixtures

src/codecs/doi/doi.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const doi2yaml = async (doi: string) =>
+    convert(doi, undefined, { from: 'doi', to: 'yaml' })
+
+  expect(await doi2yaml('10.5334/jors.182')).toMatchFile(snapshot('10.5334-jors.182.yaml'))
```
nokome: Using `jest-file-snapshot` to store snapshots in files to declutter `.test.ts` files, have syntax highlighting and still easily update snapshots using `npx jest --updateSnapshots`

src/codecs/bib/__file_snapshots__/small.yaml
```diff
@@ -0,0 +1,19 @@
...
+    type: Person
+title: A small paper
+id: articleid
+datePublished: '1924'
+isPartOf:
+  issueNumber: '9'
+  isPartOf:
+    volumeNumber: '10'
+    isPartOf:
+      title: The journal of small papers
```
beneboy: Poor Albert, reduced to writing in The journal of small papers 😢

src/codecs/crossref/index.ts
```diff
@@ -0,0 +1,52 @@
...
+      'journal-article': 'article-journal',
+      'book-chapter': 'chapter',
+      'posted-content': 'manuscript',
+      'proceedings-article': 'paper-conference'
+    }
+    csl.type = replacers[csl.type] || csl.type
+
+    return decodeCsl(csl)
+  }
+  throw new Error(`Request failed`)
```
beneboy: It's mostly personal preference but I would reverse the logic here:

```typescript
if (response.statusCode !== 200 || !response.body) {
    throw new Error(`Request failed`)
}

const data = JSON.parse(…)
```

This way the code loses one level of indentation, and I got confused when I got to the end and saw the function just ended with `throw new Error`

src/codecs/csl/index.ts
beneboy: Smart

src/codecs/orcid/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    const data = JSON.parse(response.body)
+    if (Array.isArray(data.url) && data.url.length > 0) data.url = data.url[0]
+    return coerce(data, 'Person')
+  }
+  throw new Error(`Request failed`)
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export async function encode(node: stencila.Node): Promise<vfile.VFile> {
+  throw new Error(`Encoding to an ORCID is not yet implemented`)
```
beneboy: Would it be useful to have something like Python's `NotImplementedError` to use in these cases, it might make it easier for use to locate unimplemented feature to work on?

beneboy: Plus add visibility to the end user who wouldn't be looking through our codebase and seeing `todo`s.

src/codecs/person/index.ts
```diff
@@ -2,25 +2,47 @@
...
  * Decode string data into a `Person`.
  *
  * @param data Data to decode
  */
-export function decode(data: string): Person {
+export function decodeSync(data: string): stencila.Person {
   const { name, email, url } = parseAuthor(data)
   const { title, first, middle, last, suffix } = parseFullName(name)
-  const person: Person = { type: 'Person' }
+  const person: stencila.Person = { type: 'Person' }
```
beneboy: Should we be using factory functions for things like this now? (May not be available in this branch?)

src/log.ts
```diff
@@ -24,3 +26,43 @@ export const configure = (debug: boolean = false): void => {
     }
   })
 }
+
+type CodecOperation = 'decode' | 'encode'
+
+export const errorNodeType = (
```
beneboy: I don't like the name of this function, I think it could be a little bit more verbose. Without referring back to this file it kind of loses context. Something like `logErrorNodeType` would be better.

src/codecs/person/index.ts
nokome: Yes, definitely agree we should. And yes available in this branch.

src/log.ts
nokome: Sure. If we agree this "log all errors and warnings" approach is one we want to take then we should have good names for them, because we'll be using them a lot.

beneboy: Alternatively: write code that doesn't generate errors.

src/codecs/crossref/index.ts
alex-ketch: I also wouldn't mind leaving the error handling to the `get` function, that way we don't have to do the `statusCode` etc checks each time we use it.

src/codecs/crossref/index.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export async function encode(node: stencila.Node): Promise<vfile.VFile> {
+  throw new Error(`Encoding to a Crossref query is not supported`)
```
alex-ketch: Not necessary right now, but might be nice to start considering creating our own Error types to allow different handling logic when catching the errors (ref. [custom Error handling](https://eloquentjavascript.net/08_error.html#p_o1E5pkUD5g))

src/codecs/crossref/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+  })
+  if (response.statusCode === 200 && response.body) {
+    const data = JSON.parse(response.body)
+    const csl = data.message.items[0]
+    // The output from api.crossref.org is not strictly CSL-JSON
+    // See  https://github.com/CrossRef/rest-api-doc/issues/222 for more
+    //
+    // Props `container-title`, `title`, and `ISSN` should be a string, not an array.
+    for (const prop of ['container-title', 'title', 'ISSN']) {
```
alex-ketch: More of a stylistic question than anything, but curious why the `for ... of` instead of `map`ping over the array?

src/codecs/person/index.ts
```diff
@@ -29,3 +51,31 @@ export function decode(data: string): Person {
...
+ *
+ * @param node The `Node` to encode
+ * @returns A promise that resolves to a `VFile`
+ */
+export const encode: Encode = async (
+  node: stencila.Node
+): Promise<vfile.VFile> => {
+  let content = ''
+
+  if (is<stencila.Person>('Person')(node)) {
```
alex-ketch: I can see why you added `isA` :)

src/codecs/crossref/index.ts
nokome: That was written waaaay back in April before I reached functional enlightenment :prayer_beads:

nokome:
Closed on favour of #201

PR #41, opened 2019-04-29
Adds functionality to encode end decode rPNGs

alex-ketch:
**Couple questions for this PR:**
- I wanted to discuss some of the fixture changes which were updated to match the latest definitions in the Schema repo, make sure that the new ones were indeed the way to go.
- Whether to use `#text` or `Text` for text nodes


nokome:
> Whether to use #text or Text for text nodes

As discussed, lets merge in this PR, as is, using current convention (i.e. `#text`). We'll then create sister PRs in schema and convert that will aim to get stability for the key textual types e.g. `Paragraph`, `Emphasis` etc.

I added `rpng` to the list of compilers so that I could use it from the command line but found that when I generated some rPNG files that they didn't contain the `JSON` `tEXt` chunk whereas the original test case did:

```bash
> pngcheck -tc tests/fixtures/rpng/rpng.png
File: tests/fixtures/rpng/rpng.png (1810 bytes)
Comment:
    Created with GIMP
JSON:
    {"type":"Paragraph","content":"The JSON extracted from the rPNG TEXt chunk"}
```

This was because the `unparse` function [wasn't actually calling `insert`](https://github.com/stencila/convert/pull/41/commits/7cbd6c62646155ecefde80afcb10c1d714f972e2#diff-8ae8f48db77cb406b2043836bab9b562R160). So I fixed that and in the process refactored things to (a) avoid unnecessary stringify/parse of JSON (b) have option to insert other potentially useful text chunks in the future e.g. `Software`, `Author`. See http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.Anc-text

PR #42, opened 2019-04-30
Stabilize text conversion 🐶

nokome:
This PR aims to stabilize and more fully implement converters for textual content: in particular `md`, `html`, `gdoc` and `jats`. 

### Schemas

Another aim is to help stabilize the schemas to textual nodes by informing what properties are necessary.  As such, this PR depends on a sister PR https://github.com/stencila/schema/pull/56. In `package.json` we depend on that branch: `"@stencila/schema": "github:stencila/schema#stabilize-text"`. But because the compiled `.d.ts` files are not committed in this branch, it's necesssary (any easier during active development) to link that repo after pulling this branch:

```bash
npm install
npm link @stencila/schema
```

Then, each time you make a change in the `schema` repo, build the Typescript definitions to make them available here:

```bash
npm run build:ts
```

### Pandoc

The previous version of converts for textural content were based on Pandoc (see https://github.com/stencila/convert/blob/master/src/pandoc/PandocConverter.js). I think we should look at refactoring and updating that to leverage Pandoc for `jats`, `docx`, etc conversion. Note updates Pandoc node defintions at https://github.com/jgm/pandoc-types/blob/1.19/Text/Pandoc/Definition.hs


tests/md.test.ts
```diff
@@ -87,7 +108,151 @@ First paragraph.
...
+          {
+            type: 'Verbatim',
+            value: 'verbatim'
+          },
+          ' text.'
+        ]
+      },
+      {
+        type: 'Paragraph',
+        content: ['A paragraph with ', true, ' and ', false, ' boolean values.']
```
alex-ketch: I'll admit that the conversion of `[x] == true` and `[ ] == false` caught me a bit off guard. Based on my experiences using other markdown editors I expected this to be remain unchanged, as square brackets.

nokome: Thanks for catching this and pulling me up on it. I was getting a little too far ahead and thinking about how we will represent user changeable variables (ie. inputs) in documents. The checkbox representation could be a nice repr for booleans at various places in the text (not just at the start of list items). But lets come back to that decision in a future iteration. I have removed it for now.

PR #47, opened 2019-05-07
Add commands to generate CLI binary

### LINKED ISSUE #45: Ability to package CLI as a standalone binary
To make it easy to serve and use the Convert CLI without requiring an installation of Node, we'd like to package it up as a binary package using [PKG](https://github.com/zeit/pkg)

This will be needed fairly soon to be used from Stencila Hub.

alex-ketch:
Closes #45 

Need to do some testing of the binary on other platforms.

There have been ongoing reports of [issues with Puppeteer and PKG](https://github.com/zeit/pkg/issues/204), so would be nice to verify functionalities which depend on Puppeteer (e.g. rPNG generation).

Another aspect is that the binary generation gives the following warning:
``` sh
> Warning Cannot include directory %1 into executable.
  The directory must be distributed with executable as %2.
  node_modules/puppeteer/.local-chromium
  path-to-executable/puppeteer
```

The command `npm run postpkg` copies Puppeteer/Chrome into the required folder, but it defeats the purpose of having a single binary as now we have two binaries we'd need to ship.
In addition, the puppeteer binary will likely have to change depending on the targeted platform, so adds a more overhead during the build process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 /.sandbox
+/build
 /coverage
 /dist
 /docs
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /node_modules
 /tests/**/*.to.*
 /tests/fixtures/rpng/*out.png
+/convert-deps.tgz
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/dubnium
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ check:
 
 build:
 	npm run build
+.PHONY: build
 
 docs:
 	npm run docs
--- a/package-lock.json
+++ b/package-lock.json
@@ -2194,6 +2194,24 @@
         "babel-plugin-jest-hoist": "^24.6.0"
       }
     },
+    "babel-runtime": {
+      "version": "6.26.0",
+      "resolved": "https://registry.npmjs.org/babel-runtime/-/babel-runtime-6.26.0.tgz",
+      "integrity": "sha1-llxwWGaOgrVde/4E/yM3vItWR/4=",
+      "dev": true,
+      "requires": {
+        "core-js": "^2.4.0",
+        "regenerator-runtime": "^0.11.0"
+      },
+      "dependencies": {
+        "regenerator-runtime": {
+          "version": "0.11.1",
+          "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.11.1.tgz",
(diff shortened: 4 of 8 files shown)


nokome:
This works fine, until I tried to create a rPNG:

```bash
> npm run build
> alias convert=build/stencila-convert-linux 
nokome@pluto ~/s/s/convert (45-build-binaries)> convert "<p>Hello</p>" --from html ./temp.rpng
Converting from "<p>Hello</p>" to "./temp.rpng"
converter [in] [out]

Convert between file formats

Positionals:
  in   The input file path. Defaults to standard input.                                                                          [string]
  out  The output file path. Defaults to standard output.                                                                        [string]

Options:
  --help, -h     Show help                                                                                                      [boolean]
  --version, -v  Show version                                                                                                   [boolean]
  --from         The format to convert the input from.                                                                           [string]
  --to           The format to convert the output to.                                                                            [string]

Error: Chromium revision is not downloaded. Run "npm install" or "yarn install"
    at Launcher.launch (/snapshot/convert/node_modules/puppeteer/lib/Launcher.js:119:15)
```

I get the same if I run inside the `build directory:

```bash
> cd build
> ./stencila-convert-linux "<p>Hello</p>" --from html ./temp.rpng
```

Maybe it's possible to configure Puppeteer to tell it where to look for Chromium?

I remembered where I tackled similar issues: in Nixster!
See:
- https://github.com/stencila/nixster/blob/master/src/boot.ts 
- https://github.com/stencila/nixster/blob/master/package.json

I think we need to resolve this issue in the long term so that we can distribute a single file. Also, I think we will have to reinstate the Pandoc converters in the not too distant future. It introduces a similar problem which is currently solved by lazyily downloading the binary - https://github.com/stencila/convert/blob/master/src/helpers/pandoc.js - but which may be better solved by bundling the binary into the snapshot.

nokome:
> In addition, the puppeteer binary will likely have to change depending on the targeted platform, so adds a more overhead during the build process.

I don't think we need to worry about this too much: Linux and MacOS can be put into the Travis build matrix and we have Appveyor set up for WIndows.

alex-ketch:
@nokome Mind giving this another spin when you get a chance please?

nokome:
Thanks!

```bash
> convert "<p>Yay it works</p>" --from html --to rpng ./temp.png
```

Produces:

temp

And just to confirm that's a rPNG:

```bash
> pngcheck -tc temp.png
File: temp.png (1951 bytes)
JSON:
    {
      "type": "Article",
      "content": [
        {
          "type": "Paragraph",
          "content": [
            {
              "type": "#text",
              "value": "Yay it works"
            }
          ]
        }
      ]
    }
OK: temp.png (84x53, 32-bit RGB+alpha, non-interlaced, 89.0%).
```

PR #55, opened 2019-05-09
Only build binaries for host platform

### LINKED ISSUE #52: Only build and bundle for the host platform
Currently we are building MacOS, Windows and Linux binaries all on Linux (Travis CI). We shouldn't be doing this because we are now using native binaries (Chromium and Pandoc at present). 

FWIW, the change to `pkg` options needed is:

```bash
pkg --targets=host --out-path=bin .
```

alex-ketch:
Close #52 
Thanks for providing the solution in the issue @nokome 😄


nokome:
Thanks @alex-ketch ! I've made some tweaks to simplify things (since there is only ever one sort of build - on the host). I also got the builds to work on MacOS and Windows CI's. We should be getting MacOS binaries uploaded to Github release. But some more config may be needed for Windows on Appvryor.

PR #56, opened 2019-05-10
Pandoc based converters

nokome:
This re-enables several converters that are based on Pandoc:

- `jats`
- `docx`
- `odt`
- `latex`

There is still some legacy code in `src/pandoc-bilge.js` and `src/pandoc` which we should either delete or make use of before merging this PR.

There are also plenty of `TODO`s that could be addressed or removed and turned into issues.

The main piece of functionality to be renabled is the handling of [`Table` nodes](https://github.com/stencila/convert/blob/32d012d7bbb78f922923f9083547c5d291997a3f/src/pandoc.ts#L387-L414).

You can test conversion like this:

```bash
npx ts-node --files src/cli.ts ./test.md ./test.docx
```

@alex-ketch See the `jats` converter for an example of how to specify and put a template for `docx`


src/index.ts
```diff
@@ -336,6 +342,13 @@ export async function convert(
...
   to?: string
 ): Promise<void> {
-  const node = await read(inp, from)
-  await write(node, out, to)
+  // TODO: finish refactoring this so that it handles missing
+  // inp and out args (i.e. reading from stdin, writing to stdout)
+
+  const inpFile = vfile.create({ path: inp })
+  const inpCompiler = await match(inp, from)
+  const node = await inpCompiler.parse(inpFile)
```
alex-ketch: I you mentioned this briefly during our call, but wanted to make sure we don't forget to follow up.
This change is causing the contents to always be empty for me, no matter what format/conversion I tried.

nokome: Thanks for the prod. This should now be fixed and there are additional test cases in place to catch any regressions.

src/vfile.ts
```diff
@@ -37,18 +39,66 @@ export function dump(vfile: VFile): string {
...
+ * path. This function, assesses whether a file is a file system path or not.
+ *
+ * If the string starts with `/`, `./`, `../` (or Windows compatible backslash
+ * equivalents as well as drive letter prefixed strings e.g. `C:\`)
+ * then it is assumed to be a path but the presence of the path is not checked.
+ * The reason for not checking presence
+ * here is because "if the content looks like a path then the user probably meant
+ * it to be a path". That is, if a user tries to convert the string "./file.md" then,
+ * if that file doesn't exist, she probably wants the Markdown converter to give me an error
+ * message. She probably doesn't want the `match` function to try and find some other converter
```
alex-ketch: I like that we're being gender inclusive in our writing and wanted to bring up a stylistic comment for this and our other documentation.
We could go a step further and use the [singular they](https://blog.ap.org/products-and-services/making-a-case-for-a-singular-they) in our documentation to be inclusive of gender non-binary people.

tests/index.test.ts
```diff
@@ -133,34 +116,80 @@ test('dump', async () => {
...
-    )
+  it('works with file paths', async () => {
+    expect(await read(simpleThingPath)).toEqual(simpleThing)
   })
 
   it('works with content', async () => {
     expect(await read(simpleThingJson, 'json')).toEqual(simpleThing)
   })
+
+  if (!process.env.TRAVIS)
```
alex-ketch: Out of curiosity, does this fail on Travis, or we're skipping to avoid the 10s delay on CI?
**Edit:** Noticed the other Travis check below too, which doesn't have a delay.

nokome: This fails on Travis and Appveyor https://ci.appveyor.com/project/nokome/convert/builds/24493947#L57. So this is a lazy workaround.

src/vfile.ts
nokome: :+1:

src/markdown.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Compiler for Pandoc Markdown
+ */
+
+import stencila from '@stencila/schema'
+import * as pandoc from './pandoc'
+import { VFile } from './vfile'
+
+export const mediaTypes = []
+
+export const extNames = ['markdown']
```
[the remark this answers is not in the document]
nokome: Yes, this is confusing. We have a Remark-based `md.ts` converter (that matches `.md` files implicitly due to the `mediaTypes` defined there) and a Pandoc-based `markdown.ts` converter (that  is currently not enabled because it is not included in `index.ts`.

There is a wider discussion, beyond the scope of this PR, about how/if we expose all the formats that Pandoc provides - and what to use when there is an overlap with other "native" converters like for Markdown. For now, I've remove this file.

PR #72, opened 2019-05-23
Dog-fooded and tested docs 🐕

nokome:
Stencila is all about documents, specifically reproducible, executable documents. And this repo, is all about encoding documents in alternative formats. So, shouldn't we be eating our own dog food and writing reproducible documentation using our own tools?

In addition to dog fooding, this PR is motivated by the following:

- we've encountered the issue of ensuring that examples in documentation are actually valid (see https://github.com/stencila/schema/pull/63).

- our current approach to text fixtures is a little unwieldy (see https://github.com/stencila/convert/issues/57). It would be nicer if our tests cases were more discover-able and have some narrative around them.


nokome:
To run a live-syncing development server so that you can edit files in `docs` and see the results in the browser do:

```bash
npm run docs:dev
```

PR #79, opened 2019-05-31
feat: Added logging via Stencila Logga

beneboy:
^ As per title. The other uses of console.log / console.error that haven't been replaced are because they were used to send to STDOUT used in scripts that would only be called from the command line.


nokome:
Thanks!

PR #88, opened 2019-06-05
Remove packaging

beneboy:
Remove dependency extraction, `pkg` module dependency and binary building commands.


nokome:
Thanks @beneboy. I went ahead and removed some more unnecessary stuff:

- Windows bundling script and cert
- Binary deployment in Travis config
- install script and instructions in README

All these things will need to be added into `stencila/stencila`. I'll go ahead and merge this now.

Issue #93, opened 2019-06-05
roundtrip md to docx: Unhandled Pandoc element type "SoftBreak"


jwijay:
When converting a `md` file (https://github.com/datacarpentry/R-ecology-lesson/blob/master/README.md) to a `docx`, then back to a `md`, an error gets thrown 

![Screenshot 2019-06-05 13 55 57](https://user-images.githubusercontent.com/10161095/58990057-f9ab2600-8799-11e9-8927-42d6e1f0b1cb.png)

runs successfully:
`npx ts-node --files src/cli convert examples/R-ecology-lesson/README.md ./examples/output/README-from-md.docx`

results in above error:
`npx ts-node --files src/cli convert examples/output/README-from-md.docx ./examples/output/README-from-docx.md`

PR #94, opened 2019-06-05
Doc updates + Pandoc `SoftBreak` (#93)

jwijay:
docs: Updates some outdated documentation.
fix: Resolves unhandled Pandoc `SoftBreak` in https://github.com/stencila/encoda/issues/93.


nokome:
Thanks @jwijay! I modified the code a little so that the function no longer throws and exception but instead falls back to decoding inlines to a string (as you implemented for `SoftBreak`)

Issue #90, opened 2019-06-05
md to docx: output path requires `./`


jwijay:
![Screenshot 2019-06-05 11 29 16](https://user-images.githubusercontent.com/10161095/58980535-6e279a00-8785-11e9-8971-ccc8eb92384c.png)

Probably a low-priority issue. When using encoda, the output path requires a `./`, and the error points to a missing codec rather than the path format.

e.g. passing:
`npx ts-node --files src/cli convert examples/R-ecology-lesson/README.md ./examples/output/README.docx`

but failing:
`npx ts-node --files src/cli convert examples/R-ecology-lesson/README.md examples/output/README.docx`

beneboy:
I went to fix this but the code only assumes it's a path if it starts with` `/, `./` or` ../`. Or if the file at that path exists (which the output file won't).

https://github.com/stencila/encoda/blob/0a771d4b46c9dac6fd95c81502b4a037a1b5bcb7/src/vfile.ts#L99
```
If the string starts with `/`, `./`, `../` (or Windows compatible backslash
```

I will defer to @nokome for input. On one hand I think detecting of the file could be better but also since this interface is only for devs maybe it doesn't matter. Perhaps just a message like 'if $path is a path, please prefix it with ./'

nokome:
This isn't only for devs because the same function gets used in `stencila` too. The aim is to allow the user to be able to write content or a file path at the CLI. Putting content on the CLI may not seem that that useful if you only think about doing file conversions. But for demoing and testing encodings it can be useful e.g.

```
npx ts-node --files src/cli convert "{type: 'CodeBlock'}" test.png --from json5 --to rpng
```

So the point of `isPath` is to guess if the user intended to specify a file path. I'd prefer to keep it like that.

But we need to improve it's guessing - I think by detecting if it's got an extension - like we do here: https://github.com/stencila/stencila/blob/master/src/commands/convert.ts#L79

PR #102, opened 2019-06-07
fix: Be more lenient with detecting output paths. Ref #90

beneboy:
I added an `isOuput` flag, which means `isPath` will be more lenient, since it's unlikely that we'd be passing content as the output argument on the command line (as per example in https://github.com/stencila/encoda/issues/90#issuecomment-499746559)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -164,7 +164,7 @@ export async function match(
   let extName
   let mediaType
   // If the content is a path then begin with derived values
-  if (content && vfile.isPath(content, isOutput)) {
+  if (content && (vfile.isPath(content) || isOutput)) {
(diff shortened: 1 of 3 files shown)


nokome:
Thanks @beneboy. Although this was reported for the output argument, it also affects the input argument. I tweaked your changes so that it is now possible to leave out the leading `./` on both the input and output (as long as there is a file extension).

beneboy:
@nokome ./ wasn't required on input before, since it falls back to checking if the file exists. If it doesn't we have other problems.

nokome:
@beneboy yes, you're right, brain fade on my part.

PR #104, opened 2019-06-12
Update xlsx codec

### LINKED ISSUE #63: Remove deprecated schema properties from `xlsx` converter
> :+1: This will require an change in the `xlsx` converter. https://github.com/stencila/convert/blob/master/src/xlsx.ts#L121-L161

_Originally posted by @nokome in https://github.com/stencila/schema/pull/59_

The following properties have been removed from Schemas:
- Table: `cells`
- TableCell: `position`

alex-ketch:
- Adds ability to en/decode Tables/Datatables.
- introduced `fp-ts` to the project for some type-safe utilities around ordering, function composition, etc. Can be removed/swapped out if anyone has objections.

Closes #63


src/xlsx.ts
```diff
@@ -118,53 +122,86 @@ function encodeCreativeWork(node: stencila.CreativeWork) {
...
-      let cellObject = encodeCell(cell.content[0] || null)
-      sheet[name] = cellObject
+/**
+ * Convert spreadsheet worksheet, a column-indexed table data format, to a RowMap,
+ * a row-indexed data representation.
+ *
+ * @param {xlsx.CellObject} cell
+ * @returns {stencila.TableCell}
+ */
+const worksheetToRowMap = (worksheet: Worksheet): RowMap =>
```
nokome: @param and @returns don't look right. 

Although, I'm pretty obsessive - maybe too much so - about adding doc comments to functions. I have been wondering if it's necessary to add these tags since Typescript+TypeDoc does such a good job of auto-documenting them.

Issue #103, opened 2019-06-11
Conversion of nested lists


philneff:
Conversion of documents containing lists drops lower-level items in nested lists.

Tried to convert the following list:

```
1. Fruit
    - Apple
    - Orange
    - Banana
2. Dairy
    - Milk
    - Cheese
```

Tried various conversions: Markdown to Markdown; Docx to Markdown; Markdown to Docx. In all cases converted document only included top-level list items, e.g.:

```
1.  Fruit
2.  Dairy
```

nokome:
Thanks for taking the time to submit this issue @philneff. I've created #105 to address it - hopefully we can resolve it fairly soon.

alex-ketch:
Hi @philneff 👋
I wanted to let you know that with the [v0.51.0 release](https://github.com/stencila/encoda/releases/tag/v0.51.0) we have support for nested lists. There is a currently limitation when converting to/from Google Docs, but we plan on fixing that shortly as well.
I'll leave this issue open until we have full support, but we'd love it if you could give the new version a spin meanwhile :)

nokome:
To avoid any confusion, the v0.51.0 release is of this repo, `encoda`. We are yet to update the `stencila` command line tool which Phil used at the UW eSciences workshop (still at [0.30.2](https://github.com/stencila/stencila/releases)) with this new version of `encoda`. Will ping back here when we do that.

PR #105, opened 2019-06-12
Decoding of nested lists

nokome:
This PR addresses #103. The bug arises because the decoding of `List` nodes has only been partially implemented. This is largely because I wanted to avoid adding a `ListItem` type, since it seemed somewhat redundant. However, this bug illustrates why it is necessary - to act as a container for more than one item in a node (in this case a paragraph and a (child) list).

I've just setup a test with fixtures in three formats. The test shows the proposed structure of a `ListItem` - it just has a `content` attribute (which can be `BlockContent`). This new type will have to be added to `Schema`.

@alex-ketch : more than happy if you'd like to take this on. See some of the offending lines where I took shortcuts <blush> (just taking fist child):

- https://github.com/stencila/encoda/blob/master/src/md.ts#L559
- https://github.com/stencila/encoda/blob/master/src/pandoc.ts#L498
- For `gdoc` the decoding/encoding of lists is complicated...but not impossible.


alex-ketch:
@nokome I can pick this up as I was actually briefly exploring potential solutions to this last night before I saw your comment.
My initial thought was to go look for a `ListItem` object as well, but realizing we didn't have one I did explore using a `Collection` block as a substitute.

I think long term `ListItem` makes sense and will add it to `schema` latter half of today. Let me know if you change your mind :)

nokome:
Thanks. Note that there is already https://schema.org/ListItem. 

But IMO, we should still allow for simple `List`s of nodes. In other words, we should not require that `ListItem` be used for every list item, but rather that it is used as a container for list items containing more than one content node. e.g. we should still allow for:

```yaml
type: List
items:
  - 'A string'
  - type: Emphasis
    content:
       - emphasis
```

as well as:

```yaml
type: List
items:
  - type: ListItem
    content:
       - 'A string'
  - type: ListItem
    content:
       - type: List
//...
```

Issue #34, opened 2018-10-18
xlsx to PDF conversion fails with 


apawlik:
[Reported by burque505 - Winter Laite](https://community.stenci.la/t/missing-inserted-in-alignment-preamble-xlsx-to-pdf/206)

*Even a spreadsheet with only 6 cells, each containing only text, fails with the error mentioned. The json created, however, contains all the data.(...)
EDIT:
Code problem solved with bbcode tag (trying spoiler out also). By the way, I get a useable conversion by first converting xlxs -> md and then md -> pdf, but no charts so far.*

EDIT: problematic file attached

[TM.xlsx](https://github.com/stencila/convert/files/2560813/TM.xlsx)

nokome:
Closing. The generated PDF FWIW: [test.pdf](https://github.com/stencila/encoda/files/3297549/test.pdf)

Issue #98, opened 2019-06-07
HTML: Implement `encodeDatatable` and `decodeDatatable`


nokome:
Currently, the fallback encoding of a `<stencila-thing>` is being used for a `Datatable`:

```bash
$ npx ts-node --files src/cli convert "{type: 'Datatable'}" - --from json5 --to html
<stencila-thing>{type:'Datatable'}</stencila-thing>
```

Since `Datatables` are commonly used in reproducible documents, e.g imported from CSV, it would be good to implement them soon.

PR #108, opened 2019-06-13
Add test for issue #34

nokome:
Given, recent reimplementation of `Datatable` encoding/decoding, I was hoping to close issue #34. This test is currently passing but the generated PDF doesn't look like expected because there is no HTML encoder for `Datatable`s so the fallback is being used. So, parking this as WIP until #98 is done.

The generated PDF is attached.
[34-xlsx-to-pdf.pdf](https://github.com/stencila/encoda/files/3283694/34-xlsx-to-pdf.pdf)


nokome:
Closes #34 and #98.

Issue #100, opened 2019-06-07
Puppeteer: Fix browser setup and teardown


nokome:
We have a `puppeteer.ts` module which provides an interface to Puppeteer. I recently added this so that only one browser instance is lazily instantiated by consuming codecs (e.g `pdf` and `rpng`). 

I ran into issues with the browser not closing and so, rather hurridly, added:

https://github.com/stencila/encoda/blob/master/tests/teardown.ts#L1

and

https://github.com/stencila/encoda/blob/master/src/cli.ts#L51

Neither are desirable, and the former makes the `jest --watch` exit prematurely.

I wonder if the best approach is to have one *one* browser instance (i.e. not have one per consuming codec module).

nokome:
Just to note that this also seems to make tests always pass: https://travis-ci.org/stencila/encoda/jobs/544548303#L795. Doh!

Issue #61, opened 2019-05-15
Improve speed of rPNG generation


nokome:
Currently rPNG generation is quite slow. This is not a huge issue but could become one with documents that contain many reproducible elements (e.g. lots of within text code expressions providing analysis results). The `pandoc.test.ts` which involves generation of six rPNGs is noticebly slow  https://travis-ci.org/stencila/convert/jobs/532638335#L144

### A little test

Here are Markdown document with some non-textual nodes:

```markdown
A !number(1) and a !boolean(true) and an !array(1,2,3,4,5) and an !object("a":1, "b":2)
```

When converted to Word `bin/stencila-convert temp.md ./temp.docx` the "non-textual" nodes get converted to rPNGs:

![image](https://user-images.githubusercontent.com/1152336/57753058-2c37a500-773f-11e9-9c88-56874b67a161.png)

Which takes an average of 3.5s (and shows spike in network traffic associated with fetching `highlight.js` presumably): 

```bash
bash -c "time for i in {1..10}; do bin/stencila-convert temp.md ./temp.docx; done"

real    0m35.710s
user    0m21.296s
sys     0m4.856s
```

If you remove the exclamation marks (i.e. remove the Commonmark inline extension syntax):

```markdown
A number(1) and a boolean(true) and an array(1,2,3,4,5) and an object("a":1, "b":2)
```

Then there are no rPNGs and the time drops to 1.1 seconds (only slightly above the base startup time, https://github.com/stencila/convert/issues/60).

```bash
bash -c "time for i in {1..10}; do bin/stencila-convert temp.md ./temp.docx; done"

real    0m11.092s
user    0m13.652s
sys     0m1.008s
```

### Potential solution

It seems that the first approach would to lazy, and only once, initialise a rendering page. That is, do this https://github.com/stencila/convert/blob/99975bbdcd3526d98041d98bf0f098b867d770d4/src/rpng.ts#L208-L216 outside of the `unparse` function, and only as needed.

alex-ketch:
Re. network dependency, I can investigate bundling Highlight.js as part of the package and referencing the JS that way. Without a build tool inclusion of CSS is a little tricker, since TS compiler doesn't support it natively, so we'd have to either manually bundle the CSS during build, copy-paste the CSS as a string, or introduce a build tool into the mix.
The latter option does open up further optimizations like tree-shaking mentioned in #60 though.

PR #113, opened 2019-06-16
fix: Refactor puppeteer usage to be more explicit with shutdown handling

beneboy:
Ref: #100 

In cases like this I find it better to be explicit with handling the shutdown of the browser. I have added an explicit shutdown call inside a `finally` block for rpng and pdf. 

This will probably slow down the repeated calls to open/close the browser when building RPNGs. Whether or not that will be a problem remains to be seen. If it is, we can refactor so the base Encoda caller handles the browser opening and closing and passes it into the RPNG encode function.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5139,8 +5139,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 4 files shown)


nokome:
Thanks. This essentially reverts https://github.com/stencila/encoda/pull/87 (see https://github.com/stencila/encoda/issues/61). Unfortunately, but unsurprisingly, it seems to be quite a bit slower. I have another PR incoming which does the lazy instantiation of a single browser instance properly plus the explicit shutdown of that instance in the CLI and tests. So, closing this one.

PR #136, opened 2019-07-01
Generate docs using `dir` codec

nokome:
This uses the `dir` codec to generate documentation based on `src/**/README.md` files. Currently, we use a symlink to `README.md` to provide a documentation index page. Note also that this generates a `root.json` file that can be used to generate a navigation widget.

@alex-ketch Could you merge if you have no objections. The generated HTML needs work but that will come in separate PRs here and in `thema`.


nokome:
@alex-ketch This is blocking me a little (mainly because I'd like the update docs including the README table (which is a pain to merge)). So I'm going to merge now. Feel free to revise anything in here later.

PR #163, opened 2019-07-11
JATS 'n Microdata

nokome:
The purpose of this PR is to (a) add some JATS test fixtures with real life published articles, (b) implement a better JATS codec for reading these fixtures, (c) augment the HTML codec so that it encodes structured data in a more semantic way.

So far, development of the HTML codec has focused on `Content` type nodes. These generally have a direct mapping from the JSON Schema type to a semantic HTML element. e.g `Paragraph` -> `<p>`, `TableRow` to `<tr>`. But the schema also contains types without a direct equivalent in HTML e.g.`Person`, `Article` (there is `<article>` but how do you encode properties like `datePublished`, `authors` etc). We discussed encoding these as custom elements e.g. `<stencila-person>`. This would enable us to hydrate the element into a Web Component as part of progressive enhancement.

However, there is already a standard way to do encode structured data in HTML - https://www.w3.org/TR/microdata/. This is a simple approach, consistent with our use of JSON-LD and schema.org and avoids having to namespaced custom elements.

An illustrative snippet from a HTML file generated by: 

```bash
./encoda convert src/codecs/jats/__fixtures__/elife-30274-v1.xml temp.html
```

```html
 <h1>References</h1>
<div itemprop="citation" itemtype="http://schema.org/CreativeWork"><span itemprop="author"
    itemtype="http://schema.org/Person"><span itemprop="familyName">Altman</span><span
      itemprop="givenName">DG</span></span><span itemprop="author"
    itemtype="http://schema.org/Person"><span itemprop="familyName">Royston</span><span
      itemprop="givenName">P</span></span><span itemprop="datePublished">2006</span><span
    itemprop="title">The cost of dichotomising continuous variables</span></div>
<div itemprop="citation" itemtype="http://schema.org/CreativeWork"><span itemprop="author"
    itemtype="http://schema.org/Person"><span itemprop="familyName">Biggs</span><span
      itemprop="givenName">R</span></span><span itemprop="author"
    itemtype="http://schema.org/Person"><span itemprop="familyName">Macmillan</span><span
      itemprop="givenName">RL</span></span><span itemprop="datePublished">1948</span><span
    itemprop="title">The error of the red cell count</span></div>
```

The CSS necessary to style this is verbose (would be less so with PCSS) but the structure would allow some flexibility in styling e.g.

```css
[itemtype='http://schema.org/CreativeWork'] {
  display: block;
  margin: 1em
}

[itemtype='http://schema.org/CreativeWork'] [itemprop='title'] {
  font-style: italic;
}

[itemtype='http://schema.org/CreativeWork'] [itemprop='datePublished']:before {
  content: '('
}

[itemtype='http://schema.org/CreativeWork'] [itemprop='datePublished']:after {
  content: ') '
}

[itemtype='http://schema.org/Person'] {
  display: inline
}

[itemtype='http://schema.org/Person'][itemprop='author']:after {
  content: '; '
}

[itemtype='http://schema.org/Person'] [itemprop='familyName']:after {
  content: ', '
}
```

image

We can still use a PWA approach without custom elements by hydrating Web Components from the Microdata.

Would love your thoughts @jwijay and @alex-ketch.


alex-ketch:
Overall I like this approach of using `itemprop` and `itemtype` 👍 
The details I would like to discuss, with the intent of moving us closer to semantic markup, are:
- Replace `div/span`s with, and potentially omit some of the `itemprop/itemtype` usage by swapping:
  - `itemprop="citation"` with the [`<cite>` tag](https://developer.mozilla.org/pt-BR/docs/Web/HTML/Element/cite)
  - `itemprop='title'` with a `<H3>` tag. Concern here being that we don't skip the heading level ordering
  - `temprop="datePublished"` with the [`<time>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time)

Some other interesting semantic tags we can utilize would be:
- [`<dl>`,`<dt>`,`<dd>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) tags
- [`<data>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/data) tag for encoding machine-readable values
- [`<samp>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/samp) for program output
- [`<var>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/var) for... variables :)

A medium level annoyance with using `:before`/`:after` content is that it doesn't [get included in the clipboard](https://tink.uk/accessibility-support-for-css-generated-content/) and the [accessibility is OK](https://thatdevgirl.com/blog/before-after-accessibility)(? the article is a little old now so the landscape might've changed)

nokome:
@alex-ketch, I agree we should use semantic tags where possible. My use of `<div>` and `<span>` was simply due to my haste in wanting to get a PoC implementation going. WRT to clipboard `before/after` - yes something to consider - we could add punctuation to HTML, but that would reduce flexibility for styling e.g. not having parentheses around dates.

After speaking to Adam this morning, I'm thinking we should evaluate three alternative approaches to adding semantics:

   - custom elements
   - custom classes
   - Microdata `itemtype` and `itemprop`
 
on the basis of three forms of 'readability' (which also encompasses 'authorability' i.e. something will less readability, is usually harder to write):

   - :robot: machine readability of HTML
   - :woman_technologist: human readability of HTML
   - :art: human readability of CSS

Of these, I think that human readability of HTML is the least important (because most humans will read rendered DOM and machine will generate the HTML), and machine readability of HTML is the most important (for semantic search and to be able to hydrate Web Components from HTML). Human readability of CSS is important but will need to be less often than machine reading of HTML.

### Custom elements

- e.g. `<stencila-scholarly-article>` or `<schema-org-scholarly-article>`
- :robot: :-1: has the disadvantage of being custom, not using existing standards 
- :woman_technologist: :ok_hand: quite human readable HTML
- :art: :ok_hand: OK, but a little unfamiliar for CSS authors

### Custom classes

- e.g. `<cite class="prop-citation type-scholarly-article">`
- :robot: :-1: not great, beyond use of `<cite>`
- :woman_technologist: :ok_hand: familiar use of `class` attribute
- :art: :+1: use of `class` for CSS selectors which is terse and everyone is familiar with

### Microdata

- For maximum machine readability of HTML we would use semantic elements **and** Microdata e.g. `<cite itemprop="citation" itemscope itemtype="http://schema.org/ScholarlyArticle">`
- :robot: :+1: very machine readable
- :woman_technologist: :-1: very busy for humans
- :art: :ok_hand: Although the above examples of CSS selectors are long, they could be made shorter using [custom selectors](https://cssnext.github.io/features/#custom-selectors) and/or [nesting](https://cssnext.github.io/features/#nesting):

```css
@custom-selector :--creative-work [itemtype='http://schema.org/CreativeWork']

@custom-selector :--date-published [itemprop='datePublished']

:--creative-work {
  display: block;
  margin: 1em;

  & [itemprop='title'] {
    font-style: italic;
  }

  & :--date-published:before {
    content: '('
  }

  & :--date-published:after {
    content: ') '
  }
}


@custom-selector :--person [itemtype='http://schema.org/Person']

:--person {
  display: inline;

  &[itemprop='author']:after {
    content: '; ';
  }

  [itemprop='familyName']:after {
    content: ', ';
  }
}
```

CSS custom selectors could be pre-generated from the schema. Also with the Microdata approach we could probably make the HTML encoding/decoding very declarative providing a mapping between how types/properties get represented (which elements, `itemtypes` and `itemprops` would come directly from JSON data).

alex-ketch:
@nokome I think I'm in alignment with your assignment of hierarchy of needs. Based on the emoji results, also think the 3rd (Microdata) format strikes a good balance of semantic correctness and theme "authorability".

I think any quirks in the syntax can be explained with good annotated reference stylesheets.
Sharing references for some tooling that I've created themes for in the past that have custom takes on CSS:
- [Taskpaper](https://guide.taskpaper.com/customizing-taskpaper/creating-stylesheets.html)
- [iA Writer](https://github.com/iainc/iA-Writer-Templates/blob/master/GitHub.iatemplate/Contents/Resources/github.css)
- [Atom](https://github.com/atom-community/ui-theme-template/blob/master/styles/editor.less)

nokome:
Thanks. Yes, the `stencila` style in `thema` should aim to provide that "good
annotated reference stylesheet" supported by docs in that repo (written in
`md` and encoded as HTML using `encoda` of course:).

PR #177, opened 2019-07-23
feat(TypeGuards): Remove type guards, instead import them from Schema

alex-ketch:
Relates to https://github.com/stencila/schema/pull/98

### DIFF
--- a/src/codecs/dar/index.ts
+++ b/src/codecs/dar/index.ts
@@ -71,7 +71,7 @@ export const encode: Encode = async (
   node: stencila.Node,
   options = {}
 ): Promise<vfile.VFile> => {
-  let { filePath } = options
+  const { filePath } = options
 
   const darPath = filePath || path.join(tempy.directory(), '.dar')
   await fs.ensureDir(darPath)
--- a/src/codecs/gdoc/index.ts
+++ b/src/codecs/gdoc/index.ts
@@ -102,6 +102,7 @@ class FetchToSame {
   public get(url: string): string {
     return url
   }
+
   public async resolve(): Promise<void> {}
 }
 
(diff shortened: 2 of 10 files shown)


alex-ketch:
Due to the schema inheritance changes, using `Entity` instead of `Thing` there were a couple fields on the `Link` schema the I removed. 
The changes are isolated to a commit that I can revert if we'd like to add them back to `Link`/`Entity`. What do you think @nokome?

nokome:
Ahhh yes, the `meta` issue. I added this property to `Thing` as an escape hatch - so that we could experiment with adding additional properties to any node (in Encoda, or elsewhere), without having to update `schema` and still keeping `additionalProperties: false` (in the JSON Schema). The intention with `meta` is to give tooling developers the flexibility to experiment with new properties, and then if those experiments are fruitful, to elevate those properties to being first class properties.

My first use case, as a tooling developer myself here in Encoda, was to experiment with adding `import` and `export` "directives" to `Link` nodes, so that we could do things like this in Markdown:

```markdown
This [`odt`](article-complex-ex1.out.odt){export=complex} file was generated from the simple example.
```

That commit will break that (it will just ignore the export directive). The export/import/include directives are handled in `process.ts`. I feel that this is still experimental and needs some more considered design. So I think it should be kept as experimental and therefore that those directives should stay in `meta`.

In the case of `Thing.description`, we were using that for storing link `title` attributes (from Markdown, but could also come from `<a href="...", title="...>`). So it's a different matter to `meta`.

So I suggest that:

- In schema, I move `meta` from `Thing` to `Entity` and document the above rationale for that property
- In schema, I add `title: string` to `Link`.
- you revert that commit, and make the change from using `description` to using `title`

alex-ketch:
@nokome sounds good, will revert ↩️ and update :)

nokome:
For the sake of simplicity, I'm going to make those change to Schema on master. Then `npm link` to get this and #176 working locally and then merge them both into `master`. Otherwise it feels like we're chasing our tails a bit :dog2: :poodle:

alex-ketch:
@nokome Thanks, that makes sense. I've updated this branch with the necessary changes, so feel free to merge/adjust as needed.

PR #200, opened 2019-08-01
docs(HTML Readme): Add proposed microdata details

jwijay:
Adds documentation on approach to encoding to HTML, in particular use of the Microdata https://www.w3.org/TR/microdata/

### DIFF
--- a/src/codecs/html/README.md
+++ b/src/codecs/html/README.md
@@ -108,8 +108,16 @@ Some of the main external dependencies:
         * values: `td`
         * schema (items, uniqueItems): `data`? (maybe unnecessary for styling)
         * name: `th`?
+* SoftwareSourceCode:
+    * codeRepository: `a href`
+    * codeSampleType: `data`?
+    * maintainers: `ol` of `span` Person/Organization itemtypes
+    * programmingLanguage: `data` or `span` (maybe unnecessary for styling)
+    * runtimePlatform: `ol` of `data` or `span`
+    * softwareRequirements: `ol` of `span` SoftwareSourceCode/SoftwareApplication
+    * targetProducts: `ol` of `span` SoftwareApplication
 
-### Schema nodes we won't be adding microdata to
+### Schema nodes with direct mapping to semantic tags
 * Delete: `del`
 * Emphasis: `em`
 * Heading: 
@@ -119,8 +127,37 @@ Some of the main external dependencies:
 * Link: `a`
     * content
     * target => href
-* List: `ol`, `ul` (with `input type="checkbox"` option)
+* List: `ol`, `ul` 
+  * ListItem => `li` (with `li` > `input type="checkbox"` option)
 * Quote
-* Strong
-* Subscript
-* Superscript
+* Strong => `strong`
+* Subscript => `sub`
+* Superscript => `sup`
+* Table => `table`. `table` > `tr`, or `table` > `thead`/`tfoot` > `tr`
+    * TableRow => `tr`.
+        * `kind`: header => `thead` > `tr`
+        * `kind`: footer => `tfoot` > `tr`
+    * TableCell => `th`, `td`
+        * `colspan`, `rowspan`
+        * `kind`: header => `th`
+        * `kind`: data => `td`
+* ThematicBreak: `hr`? (not sure if we should consider other options, such as splitting nodes by ThematicBreaks and grouping them into `section` tags instead)
+* VideoObject: `video`
+    * caption => `track`
+    * thumbnail => poster (attribute)
+    * transcript => `track`? (not sure if different from caption)
+
+### Schema nodes we won't be adding microdata to
+* BlockContent
+* CodeBlock?
+* CreativeWorkTypes
+* Entity
+* Environment?
+* Include?
+    * source
+    * mediaType
+    * hash
+    * content
+* InlineContent
+* Mark
+* Thing
--- a/src/codecs/html/README.md
+++ b/src/codecs/html/README.md
@@ -45,16 +45,11 @@ Some of the main external dependencies:
 
 ### Schema nodes to add microdata to
 * Article
-    * `div` `itemtype='schema:Article'`
+    * `article` `itemtype='schema:Article'`
     * itemprops:
         * authors* (author?): `ol`
         * title*: `h1`? (+ `title`)
         * environment: `data`?
-* Audio
-    * `audio`
-    * itemprops:
-        * caption*: `figcaption`
-        * transcript
 * Brand
     * `picture`/`img` `itemtype='schema:Brand'`
     * itemprops:
@@ -62,22 +57,23 @@ Some of the main external dependencies:
         * logo*: `source`/`image`
         * reviews
 * Code
-    * `div` `itemtype='stencila:Code'`
+    * `code` `itemtype='stencila:Code'`
     * itemprops:
         * language* (maybe important for setting syntax highlighting): `data`
-        * value*: `code`
+        * value*: set as `code` textContent
 * CodeChunk
-    * `div` `itemtype='stencila:CodeChunk'`
+    * `<pre><code>...</code></pre>` `itemtype='stencila:CodeChunk'`
     * itemprops:
-        * outputs: `ol` of `samp`/`output` (possibly include `figure`/`table`)
+        * outputs: `pre` containing `samp`/`output` (see https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element)
+            * In future, consider including `figure`/`table`
 * CodeExpression
-    * `div` `itemtype='stencila:CodeExpr'`
+    * `code` `itemtype='stencila:CodeExpr'`
     * itemprops:
-        * value:`code` (syntax highlighting from Code node's language prop)
+        * value: set as `code` textContent (syntax highlighting from SoftwareSourceCode node's programmingLanguage prop)
 * Collection
-    * `nav` `itemtype='schema:Collection'`
+    * `ol` `itemtype='schema:Collection'`
     * itemprops:
-        * parts: `ol` of CreativeWorks/Articles (just title, author, and URIs/relative links?)
+        * parts: `li > a href` CreativeWorks/Articles (just title, author, and URIs/relative links?)
 * ContactPoint
     * `address` `itemtype='schema:contactPoint'`
     * itemprops:
@@ -118,6 +114,9 @@ Some of the main external dependencies:
     * targetProducts: `ol` of `span` SoftwareApplication
 
 ### Schema nodes with direct mapping to semantic tags
+* AudioObject: `audio`
+    * name: `figcaption`, or `p`
+    * caption, transcript => `track` (if `video`), or `data`. [Consider using `video` instead of `audio` to enable captions](https://www.iandevlin.com/blog/2015/12/html5/webvtt-and-audio/), or display separately as `div`.
 * Delete: `del`
 * Emphasis: `em`
 * Heading: 
@@ -145,7 +144,7 @@ Some of the main external dependencies:
 * VideoObject: `video`
     * caption => `track`
     * thumbnail => poster (attribute)
-    * transcript => `track`? (not sure if different from caption)
+    * transcript => `track`?
 
 ### Schema nodes we won't be adding microdata to
 * BlockContent
--- a/src/codecs/html/README.md
+++ b/src/codecs/html/README.md
@@ -56,11 +56,27 @@ Some of the main external dependencies:
         * name: maybe unnecessary, just use in alt tag?
         * logo*: `source`/`image`
         * reviews
+* Cite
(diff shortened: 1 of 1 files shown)


src/codecs/html/README.md
```diff
@@ -38,3 +38,126 @@ Some of the main external dependencies:
...
 
 - `json5` for more human readable representation of arbitrary values.
+
+---
+
+## Microdata and semantic tag usage (WIP)
+
+### Schema nodes to add microdata to
+* Article
+    * `div` `itemtype='schema:Article'`
```
nokome: What about `<article itemtype="schema:Article">`?

src/codecs/html/README.md
```diff
@@ -38,3 +38,126 @@ Some of the main external dependencies:
...
+    * itemprops:
+        * authors* (author?): `ol`
+        * title*: `h1`? (+ `title`)
+        * environment: `data`?
+* Audio
+    * `audio`
+    * itemprops:
+        * caption*: `figcaption`
+        * transcript
```
alex-ketch: Right now both `caption` and `transcript` are simple strings. Something to evaluate would be to:
- either make `transcript` be a nested `CreativeWork` or an `Article`
- change `caption` conform to a `uri string` and reference a `.vtt` file. This way we can use the HTML [`<track> tag`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track). Reason I say this is that I don't think I'd want to avoid coupling the `audio` tag to always be inside a `figure` element

src/codecs/html/README.md
```diff
@@ -38,3 +38,126 @@ Some of the main external dependencies:
...
+* Brand
+    * `picture`/`img` `itemtype='schema:Brand'`
+    * itemprops:
+        * name: maybe unnecessary, just use in alt tag?
+        * logo*: `source`/`image`
+        * reviews
+* Code
+    * `div` `itemtype='stencila:Code'`
```
alex-ketch: I'd propose using the [`<code> tag`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code)

src/codecs/html/README.md
```diff
@@ -38,3 +38,126 @@ Some of the main external dependencies:
...
+    * itemprops:
+        * language* (maybe important for setting syntax highlighting): `data`
+        * value*: `code`
```
alex-ketch: The value would be the [`textContent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) of the `<code>` tag

src/codecs/html/README.md
```diff
@@ -38,3 +38,126 @@ Some of the main external dependencies:
...
+* CodeChunk
+    * `div` `itemtype='stencila:CodeChunk'`
```
alex-ketch: According to the spec the [recommended way](https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element) for code blocks is to use a `<pre><code>...</code></pre>`

Also note the second example in this sections which shows the usage of the `<samp>` tag for outputs

src/codecs/html/README.md
```diff
@@ -38,3 +38,126 @@ Some of the main external dependencies:
...
+    * itemprops:
+        * outputs: `ol` of `samp`/`output` (possibly include `figure`/`table`)
+* CodeExpression
+    * `div` `itemtype='stencila:CodeExpr'`
```
alex-ketch: Same here as for `Code` schema, I think we can use the `<code> tag`

src/codecs/html/README.md
```diff
@@ -38,3 +38,126 @@ Some of the main external dependencies:
...
+    * itemprops:
+        * value:`code` (syntax highlighting from Code node's language prop)
+* Collection
+    * `nav` `itemtype='schema:Collection'`
```
alex-ketch: I'm a little hesitant to make the assumption that all `Collections` are navigable collections. Using either `ol` or `ul` seems more natural to me, that way the user can decide when and where to wrap in a `nav` tag as needed

src/codecs/html/README.md
```diff
@@ -38,3 +38,126 @@ Some of the main external dependencies:
...
+* Superscript => `sup`
+* Table => `table`. `table` > `tr`, or `table` > `thead`/`tfoot` > `tr`
+    * TableRow => `tr`.
+        * `kind`: header => `thead` > `tr`
+        * `kind`: footer => `tfoot` > `tr`
+    * TableCell => `th`, `td`
+        * `colspan`, `rowspan`
+        * `kind`: header => `th`
+        * `kind`: data => `td`
+* ThematicBreak: `hr`? (not sure if we should consider other options, such as splitting nodes by ThematicBreaks and grouping them into `section` tags instead)
```
alex-ketch: Yeah, kinda I like the idea of wrapping in `section` tags 👍
Though it does raise questions about decoding `section` tags and the extra logic we'd need to make sure we don't keep nesting `sections` when encoding multiple times as well as the different ways people can (ab)use them.

src/codecs/html/README.md
```diff
@@ -38,3 +38,126 @@ Some of the main external dependencies:
...
+        * datePublished/date: `span time`
+        * editors: `ol` of `span` Person itemtypes
+        * funders: `ol` of `span` Person/Organization itemtypes
+        * isPartOf: `data`?
+        * licenses: `ol` of `a href`
+        * parts/hasPart: `ol` of figure/table/object/URI (maybe unnecessary for styling)
+        * publisher: `span` Person/Organization itemtype
+        * text: `data`? (maybe unnecessary for styling)
+        * title/headline: `h1`
+        * version: `span`
```
alex-ketch: This isn't related to this specific schema, but thinking of including all these properties in the output seems a bit overwhelming to me. Not sure how we handle going about laying all this information out in a useful way without being simply cluttered and noisy 😬

src/codecs/html/README.md
alex-ketch: Agreed on using `article` but this ties back to your overall comment:

> Perhaps a useful guideline would be to use itemtype on any schema:Thing, even if they have HTML equivalents. e.g. use <video itemtype="http://schema.org/VideoObject">. I think that for machine-readability, that is useful.

I would argue that always adding `itemtypes` might be redundant, and I'd prefer not to commit to things if we're not actively making use of them as such things easily gather dust and add to responsibilities.
The semantic tags themselves are already machine readable and the tags would be adding to the page download sizes.

src/codecs/html/README.md
jwijay: I think I may have confused the `caption` and `name` properties for AudioObject.

- Re: `figcaption` usage, I was looking at [this example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio). Maybe instead of `caption` it should be `name` that gets encoded as a `figcaption`. Or we could use a `p`, or preserve the name in `data`.
- Re: `track` usage, I don't believe we can add captions/transcripts the same way in `audio` as we can in `video`. This may be a bit in the weeds, but a workaround is to [use `video` tags instead of `audio](https://www.iandevlin.com/blog/2015/12/html5/webvtt-and-audio/), or display the captions/transcripts separately.
- Both those changes make sense: `caption` conforming to a `uri string`/referencing a `.vtt` file, and changing transcript to a nested `CreativeWork`/`Article`.

src/codecs/html/README.md
jwijay: Oh! I think this was the redundant example I was trying to explain yesterday. It slipped my mind that we could set props to the encoded tag's contents instead of storing it as another tag with itemprop. 🤦‍♀ I'll go ahead and update this one.

src/codecs/html/README.md
nokome: Fair point. I think we need to find out more about the "consumers" and what they want - in this case the web crawlers / other machines. If they semantically link a HTML `<article>` to a https://schema.org/Article, then yes it's redundant. If not, then it adding these `itemtype` provides that semantic link for them. Note that I am not proposing using `itemprop` on all nodes, only those derived from `Thing`, which will exclude many of the content nodes, like `h1`, `p` etc.

src/codecs/html/README.md
jwijay: Yeah, that might make things a bit more complicated. Maybe we'd want to include `itemtype='stencila:ThematicBreak'` either in the pre-break `<section>` tag, or in an `<hr>` in between `section` tags (and ignore `section` tags if they're previous/next siblings). Definitely seems simpler, but less semantic to stick with just `hr`.

nokome:
In this design process, I wonder if we should consider the needs of two sets of consumers:

1. Machines - mostly traditional web crawlers, but maybe also more semantically aware tools e.g. https://www.semanticscholar.org/

2. Designers - humans writing CSS

We are aiming to design an approach that can simultaneously meet the needs of both sets of consumers - semantically rich HTML for machines and semantically oriented targets for CSS selectors. So perhaps our design choices should be assessed against both of these.

To learn more about the needs of the machines, I think we need to reach out to people like Dan Brickley (Schema.org, Microdata etc Developer Advocate at Google) and others. e.g. Google recommends using JSON-LD in the `<head>` over MIcrodata (but we don't want to repeat the entire document there, and we want the Microdata attributes for CSS selectors).

For the needs of the designers, I think we should start by eating our own dog food and see how well/easily we can achieve acceptable styles using this approach and "semantic" CSS selectors: https://github.com/stencila/encoda/pull/163#issuecomment-510728508

src/codecs/html/README.md
beneboy: Just use CSS to make it look good 🤷‍♂️

src/codecs/html/README.md
```diff
@@ -38,3 +38,218 @@ Some of the main external dependencies:
...
+        * publisher: `span` Person `name` (or combined `givenNames`+`familyNames`)/Organization `legalName`.
+        * text: `data`? (maybe unnecessary for styling)
+        * title/headline: `h1`
+        * version: `span`
+* DataTable:
+    * `table` or `figure table` `itemtype='stencila:Datatable'`
+    * columns: probably want to convert these DatatableColumns into rows when encoding into HTML, so they're made up of `tr` and `td`/`th`. Maybe keep as `data`?
+        * values: `td`
+        * schema (items, uniqueItems): `data`? (maybe unnecessary for styling)
+        * name: `th`?
```
beneboy: DataTable -> Columns -> Rows is the opposite of `table` -> `tr` -> `td`

src/codecs/html/README.md
```diff
@@ -38,3 +38,218 @@ Some of the main external dependencies:
...
+        * isPartOf: `data`?
+        * licenses: `ol` of `a href`
+        * parts/hasPart: `ol` of figure/table/object/URI (maybe unnecessary for styling)
+        * publisher: `span` Person `name` (or combined `givenNames`+`familyNames`)/Organization `legalName`.
+        * text: `data`? (maybe unnecessary for styling)
+        * title/headline: `h1`
+        * version: `span`
+* DataTable:
+    * `table` or `figure table` `itemtype='stencila:Datatable'`
+    * columns: probably want to convert these DatatableColumns into rows when encoding into HTML, so they're made up of `tr` and `td`/`th`. Maybe keep as `data`?
```
beneboy: DataTable -> Columns -> Rows is the opposite of `table` -> `tr` -> `td`

src/codecs/html/README.md
```diff
@@ -38,3 +38,218 @@ Some of the main external dependencies:
...
+        * affiliations: `ul` of `span` Organization `legalName` + `departments`
+        * emails: `data`? (maybe unnecessary for styling)
+        * familyNames: `ul` of `span`
+        * funders: `data`? (maybe unnecessary for styling)
+        * givenNames: `ul` of `span`
+        * honorificPrefix: `span`
+        * honorificSuffix: `span`
+        * jobTitle: `span`
+        * memberOf: `ul` of `span` Organization `legalName` + `departments`, or `data`? (maybe unnecessary for styling)
+        * telephoneNumbers: `data`? (maybe unnecessary for styling)
```
beneboy: `ul` of `a` like this `<a href="tel:8675309">`?

jwijay: Do you think we need to display phone numbers when including a `Person` in a reproducible document/article?

It looked like the main use case for `Person` is to display lists of author names, publishers, editors, etc. For `Organization`, I think it makes sense to display phone numbers. I'm not sure if authors/individuals would want their phone numbers publicly available.

beneboy: I have no opinion one way or the other, just pointing out the use of tel: attribute.

nokome:
In case it helps, I just wanted to point out that the schema.org docs provide examples of Microdata for most of the types e.g. scroll to the bottom of https://schema.org/Person and click on one of the 'Microdata' tabs.

nokome:
Have merged this into the related, but newer, https://github.com/stencila/encoda/pull/236. So closing this one now.

PR #202, opened 2019-08-04
Refactor HTML codec

alex-ketch:
This is still WIP, but sharing direction that I ended up taking.

Tried using the standard we decided to use "`const` function expression" but that didn't work out due to functions not being able defined at call locations. The original intent for this style was to be able to statically type check that the codec functions (mainly `encode` and `decode`) conform to the same signature.
The best way I found to do it was to use classes.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,12 @@
 {
   "extends": "@stencila/eslint-config",
   "rules": {
+    "no-use-before-define": "off",
+    "@typescript-eslint/no-use-before-define": [
+      "error",
+      { "functions": false, "typedefs": false }
+    ],
+    "no-shadow": ["error", { "builtinGlobals": false, "hoist": "functions" }],
     "@typescript-eslint/strict-boolean-expressions": 0
   }
 }
--- a/src/codecs/html/index.ts
+++ b/src/codecs/html/index.ts
@@ -4,7 +4,13 @@
 
 import { getLogger } from '@stencila/logga'
 import stencila from '@stencila/schema'
-import { isArticle, nodeType, markTypes } from '@stencila/schema/dist/util'
+import {
+  isArticle,
+  isBlockContent,
+  isInlineContent,
+  markTypes,
+  nodeType
+} from '@stencila/schema/dist/util'
 import collapse from 'collapse-whitespace'
 import escape from 'escape-html'
 import fs from 'fs'
@@ -16,12 +22,13 @@ import { html as beautifyHtml } from 'js-beautify'
 import jsdom from 'jsdom'
 import JSON5 from 'json5'
 import path from 'path'
-import { Encode, EncodeOptions } from '../..'
+import { EncodeOptions } from '../..'
 import { columnIndexToName } from '../../codecs/xlsx'
 import bundle from '../../util/bundle'
 import * as vfile from '../../util/vfile'
+import { Codec } from '../types'
 
-const document = new jsdom.JSDOM().window.document
+const doc = new jsdom.JSDOM().window.document
 
 const log = getLogger('encoda:html')
 
@@ -29,22 +36,6 @@ const log = getLogger('encoda:html')
 const slugger = new GithubSlugger()
 
 export const mediaTypes = ['text/html']
-
-/**
- * Decode a `VFile` with HTML contents to a `stencila.Node`.
- *
- * @param file The `VFile` to decode
- * @returns A promise that resolves to a `stencila.Node`
- */
-export async function decode(file: vfile.VFile): Promise<stencila.Node> {
-  const html = await vfile.dump(file)
-  const dom = new jsdom.JSDOM(html)
-  const document = dom.window.document
-  collapse(document)
-  const node = decodeNode(document)
-  if (!node) throw new Error(`Unable to decode HTML`)
-  return node
-}
 
 export const beautify = (html: string): string =>
   /* eslint-disable @typescript-eslint/camelcase */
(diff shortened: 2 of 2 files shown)


nokome:
I don't understand what the benefits of refactoring to classes are. It's not that I'm opposed to it, it's just that we need to weigh the benefits against the costs of churn and time involved in refactoring all 31 codecs to using classes. 

I've spent a fair amount of time refactoring away from classes (https://github.com/stencila/encoda/tree/v0.30.0/src) and have assiduously been avoiding them since (even when that means having to pass state between decoding/encoding functions e.g. https://github.com/stencila/encoda/blob/v0.61.2/src/codecs/jats/index.ts#L543).

Here's a quick shot at it...

Advantages:
- it feels a little more natural to describe a codec as a class that implements an interface (with `decode` and `encode` functions) than as a module that implements an interface
- if a particular code needs to keep track of state, they can instantiate an instance and use `this` instead of having to pass a state variable around

Disadvantages:
- needing to prefix functions with `HtmlCodec.` or `this`

Costs:
- need to refactor all codecs and `src/index.ts` to use classes (?)

alex-ketch:
@nokome you raise good points, and I'm also hesitant to increase churn.
To elaborate a but further on my reasoning for going with classes, was to make use of the safety guarantees provided by static typings. I personally tend to avoid using classes if at all possible, but I wasn't able to find an alternate solution.

We make frequent use of dynamic function calls/matches ([one example here](https://github.com/stencila/encoda/blob/2d06f6d4dbe4e658f798827310784e52e078c891/src/index.ts#L202-L220)), and we lose some of the value that TypeScript brings to projects.
As I mentioned in the original post, one of the primary motivating factors for me was to allow certain level of type safety by ensuring that the various codecs all follow some sort of shared interface.
The problem I see is that:
- We make a lot of recursive and out of order function calls
- function declarations and their ability to be hoisted are needed to allow for this type dynamic call order
- TypeScript as far as I'm aware doesn't allow for declaring a shared type signature of a function declaration

Given the above Classes was the solution I went with.

> needing to prefix functions with HtmlCodec. or this

If we'd like we can keep the `encodeX`/`decodeX` calls out of the class, but I can also envision a way for us to statically know feature completion of each codec or provide warnings when implementing new nodes.

nokome:
> to allow certain level of type safety by ensuring that the various codecs all follow some sort of shared interface

I thought that the `Codec` interface already provided that?

https://github.com/stencila/encoda/blob/2d06f6d4dbe4e658f798827310784e52e078c891/src/index.ts#L65-L75

I don't understand how using a class for that interface provides any further type guarantees. Is it because if you use `implements` or `extends` then the compiler will complain if you don't have consistent function implementations?

> - We make a lot of recursive and out of order function calls
> - function declarations and their ability to be hoisted are needed to allow for this type dynamic call order
> - TypeScript as far as I'm aware doesn't allow for declaring a shared type signature of a function declaration

I'm afraid that after several reads I still don't understand much of this - what the problems are and how using classes solves it. For example, I thought that `Encode` is a share type signature for a function:

https://github.com/stencila/encoda/blob/2d06f6d4dbe4e658f798827310784e52e078c891/src/index.ts#L125

I'm still not opposed to using classes (in fact if it wasn't for the refactoring work involved I think I favour them because they seem a better fit for the notion of a "codec") but I'm struggling to understand the type safety advantages.

In any case I wonder if the refactoring could be done pretty easily by modifying `match()` etc in `indext.ts` and simply adding a class that wraps the existing module-based codecs:

```ts
import * as jats from '.'
export class JatsCodec implements Codec {
  public static mediaTypes = jats.mediaTypes

  public static extNames = jats.extNames

  public encode = async (
    node: stencila.Node,
    options: EncodeOptions = {}
  ): Promise<vfile.VFile> => {
    return jats.encode(node, options)
  }
}
```

Although that seems a little pointless, which maybe reflects my lack of understanding between class-based codecs and module-based codecs.

alex-ketch:
> I thought that the `Codec` interface already provided that?

That's true, but in the current state of things we have no way of making the individual codecs conform to this interface. For example I can go into a codec and delete the `decode` function without any warnings.
Using "find references" shows that the `Codec` interface is only used as the return value of the [`match` function](https://github.com/stencila/encoda/blob/256eaf522e97ddea2d526a4557bd698c30cba47a/src/index.ts#L148-L152).

If you look at the inferred values of the returned values inside that function, they're all `any`<sup>1</sup> that get cast to `Codec` because of the return type signature of the `match` function. Meaning that there is no way to know at compile time if the codec actually is a valid codec with both `encode`/`decode` functions and properties (`mediaTypes`, etc).
That's what I meant that it's breaks the promises of static typings, and was something I wasn't able to solve without classes (or faux-classes of exporting an object with those properties).

> In any case I wonder if the refactoring could be done pretty easily by modifying match() etc in indext.ts and simply adding a class that wraps the existing module-based codecs:

Yup, that's along the same lines of what I suggested in the last paragraph of my previous comment: "If we'd like we can keep the encodeX/decodeX calls out of the class, …"

Meaning we'd have a the top-level `encode` and `decode` methods in the exported class, while keeping the `encodeHeading/decodeArticle/etc` as simple functions. We can take this approach initially, to keep changes minimal, while improving type safety.
Then down the road evaluate benefits/viability of having static type-checking around implementation parity between Schema nodes and individual codec's ability to handle them.

---

- <sup>1</sup> This is the result of the previously mentioned use of dynamic function calls/matches [like here](https://github.com/stencila/encoda/blob/2d06f6d4dbe4e658f798827310784e52e078c891/src/index.ts#L203)

nokome:
Thanks @alex-ketch for your patience in explaining this, I understand better now.

I'm in favour of doing a refactoring to codecs being classes given that, based on this discussion, it seems that the first pass refactoring could be done in less than a day. Obviously this will need to be done simultaneously across all codecs. Could you go ahead and do that.

PR #211, opened 2019-08-05
chore: Refs and factory function refactoring

beneboy:
WIP: Add use of PublicationVolume and PublicationIssue for references,
moving to factory functions.


beneboy:
This requires Figure support from schema before merging.

PR #217, opened 2019-08-09
Refactor codecs to classes

alex-ketch:
Refactors codecs to use JS Classes, with the intent being to have more type safety and guarantees that the codecs implement all the necessary methods and that the type signatures are standardized.

There are a lot of lines changes, but they mainly follow the pattern of:

```diff
- export decode...
- export encode...
+ export class MyCodec extends Codec implements Codec {
+  decode =  ...
+  encode =  ...
+ }
```

I had to [update a bunch of snapshots](https://github.com/stencila/encoda/pull/217/commits/33ddfb46c443e073bd6086d236ae43b2cffa7d6d) to get the tests to pass, but not entirely sure why this was the case. Would appreciate some sanity check on this please.


src/codecs/__mocks__/ssf.ts
```diff
@@ -4,10 +4,20 @@
...
+import { Codec } from '../types'
 
 export const fileNames = ['super-special-file']
 export const extNames = ['ssf']
 export const mediaTypes = ['application/vnd.super-corp.super-special-file']
 export const sniff = async (content: string) => /^SSF:/.test(content)
 export const decode = async (file: VFile) => null
 export const encode = async (node: Node, options: {} = {}) => create()
+
+export class Ssf extends Codec implements Codec {
```
beneboy: Why do all the codecs both `extend` and `implement` `Codec`? Shouldn't just extending accomplish the method checking?

src/codecs/doi/index.ts
```diff
@@ -3,29 +3,34 @@
...
+    if (!match) throw new Error('Unable to parse content')
+    const doi = vfile.load(match[4])
+    return Doi.csl.decode(doi, { format: '@doi/id' })
+  }
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-export async function encode(node: stencila.Node): Promise<vfile.VFile> {
-  throw new Error(`Unparsing to DOI is not yet implemented`)
+  public encode = async (): Promise<vfile.VFile> => {
+    throw new Error(`Unparsing to DOI is not yet implemented`)
```
beneboy: I realise this is not really part of this refactoring but this message should be `encoding to DOI…`

src/codecs/html/html.test.ts
```diff
@@ -1,16 +1,17 @@
 import stencila from '@stencila/schema'
 import fs from 'fs'
 import { dump, load } from '../../util/vfile'
-import { decode, encode } from './'
-import { EncodeOptions } from '../..'
+import { HTML } from './'
+
+const { encode, decode } = new HTML()
```
beneboy: I know @nokome and I are kind of on the opposite ends of the code-naming-verbosity spectrum but having a class just called `HTML` kind of feels like it could get confusing. Not so much with the other codecs but something dealing with HTML is quite common in the Javascript world. Just pointing this out, I don't mind if we keep or change it.

src/codecs/jats/index.ts
```diff
@@ -3,96 +3,89 @@
  */
```
beneboy: Not sure if you rebased before refactoring this file, seems to have lost some of the changes from the `jats-refactor` branch.

src/codecs/types.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export abstract class Codec<
+  EncodeOptions extends object = {},
+  DecodeOptions extends object = {}
+> {
+  /**
+   * An array of [IANA Media Type](https://www.iana.org/assignments/media-types/media-types.xhtml)
+   * that the codec can decode/encode.
+   */
+  public abstract mediaTypes: string[]
```
beneboy: Can this (and other properties) be marked `readonly`?

src/codecs/html/html.test.ts
nokome: I'd be happy to add `Codec` to all the class names e.g. `HtmlCodec` (actually, I think I prefer that :)

src/codecs/jats/index.ts
nokome: Yes, I think that has something to do with the need to update a lot of the JATS test snapshats.

src/codecs/__mocks__/ssf.ts
alex-ketch: It has to do with how the implementation of properties are enforced and the ability to check class inheritance at runtime.

Screen Shot 2019-08-12 at 13 19 03

src/codecs/jats/index.ts
alex-ketch: hmmm, thought I was careful not to mangle anything, but clearly I was mistaken. Will give it another go

src/codecs/types.ts
alex-ketch: @beneboy unfortunately the answer is no :/
Discussion tracking the issue here: https://github.com/Microsoft/TypeScript/issues/29634

For now I've manually set the `readonly` prop in the derived classes

alex-ketch: @beneboy unfortunately the answer is no :/
Discussion tracking the issue here: https://github.com/Microsoft/TypeScript/issues/29634

For now I've manually set the `readonly` prop in the derived classes

src/codecs/doi/index.ts
alex-ketch: Fixed in https://github.com/stencila/encoda/pull/217/commits/0bd50decfa065648d1188e1adc2fa7dea5da002e

src/codecs/__mocks__/ssf.ts
nokome: Thanks for detailed explanation @alex-ketch - if you still have those demo snippets available it might be worth putting them in a `docs/developing-a-codec.md`:

```markdown
## FAQ

....

#### Why do all the codecs both extend and implement Codec? Shouldn't just extending accomplish the method checking?

It has to do with how the implementation of properties are enforced and the ability to check class inheritance at runtime.

....

```

alex-ketch: Good idea @nokome
Documented in https://github.com/stencila/encoda/pull/217/commits/88d6152ed498239b9f0ef03cc1cab311056d89ea

alex-ketch:
This should be good to merge now I think

beneboy:
> This should be good to merge now I think

Looks G to M

PR #230, opened 2019-08-19
Person: detect and use ORCID, add docs

nokome:
Some upgrades and docs for the `person` codec inspired by / following on from recent work on `ipynb` codec.


src/codecs/person/README.md
```diff
@@ -1 +1,145 @@
...
+  ],
+  "type": "Person"
+}
+```
+
+> ⚠️ The `orcid` codec is work in progress and when complete willl return more metadata than shown above.
+
+### Text parsing
+
+If no ORCID is detected, or if ORCID decoding fails (e.g. due to no network access) then the text string is parsed to extract:
```
beneboy: Not _specifically_ related to this but I wonder if it would be useful to run Encoda with an option to disable network functions. For this function it would fall back to string parsing. Other functions that require network would fail with an error.

Reasons for offering this: privacy for users, speed up codecs.

nokome: That seems like a worthwhile feature, not least of all to allow for testing that Encoda is usable without a connection. Since we are using `nock` already, the easiest way to implement that may be to simply run `nock.disableNetConnect()` if the `--nonet` option is supplied. Only issue there is that it may interfere with `puppeteer` - although `puppeteer.launch({ pipe: true })` may provide a workaround.

PR #233, opened 2019-08-21
Add more node types & extract more data from HTML nodes

### LINKED ISSUE #216: HTML: implement encoding for more node types
With recent updates to the JATS codec, we are able to generate HTML for complete, real-life examples of research articles. Amongst other uses, these can be used as examples for testing styles in Thema. Attached is example HTML generated from one of the JATS fixtures using:

```
./encoda convert src/codecs/jats/__fixtures__/elife-30274-v1/main.jats.xml temp.html
```

There are several, new node types that are currently getting encoded using the default fallback encoding of a `<stencila-thing>` for which custom encode functions need to be implemented for:

- [ ] `Cite`
- [ ] `CiteGroup` (although I don't think the JATS codec is currently generating those)
- [ ] `Figure`
- [ ] `Collection` (in this context being used for a figure group)

[temp.html.zip](https://github.com/stencila/encoda/files/3484958/temp.html.zip)

alex-ketch:
Closes #216 and implements transforming more data surrounding references and citations (this includes `Person`, `Organization`, `CreativeWork`, and a few other nodes)

@jwijay Please note that some of the HTML structure you were working with in Thema might not line up 100% with the results of Encoda, so it's something we should review together. I tried to follow the outline detailed in [your PR](https://github.com/stencila/encoda/pull/200), while keeping the output as lean and semantic as possible. However I think there were differences when compared to the test HTML found in the [Thema PR](https://github.com/stencila/thema/pull/10).

This PR:
- Adds support for Cite, CiteGroup, Figure, Collection nodes
- Handles more props on Article/Person/Org/Ref/CreativeWork nodes (both Stencila Schema and HTML elements)
- Fixes [a failing test](https://travis-ci.org/stencila/encoda/jobs/574556462#L578-L583) due to [unhandled `undefined` value](https://github.com/stencila/encoda/commit/2388d7b2be6dab6ad754c1cd7cdce56845feb68a#diff-eed3108d015c259115fb4f93110337cc)
- Fix test failures due to false TypeScript errors by [adding a `tsconfig.prod.json` file](https://github.com/stencila/encoda/pull/233/commits/a9fe1f81bbe85245f2e12bc0b4e68f82b748643d). The default config file is used by default, but the new prod config is used when running `npm run build`. This fixes errors which would show up when running tests locally. Namely complaints about the Jest matchers not being defined (see screenshot below)
Screen Shot 2019-08-21 at 16 03 46
- Adds a new custom Jest matcher called `toEqualStringContent` which compares two strings, disregarding the whitespace. Useful for comparing HTML outputs without caring about indentation

I've also identified a couple bugs/contradictions in the test-cases and will address them separately.

Has to do with the definition of `Inline` and `BlockContent` schemas and whether:
- `BlockQuote` content can be a `string (InlineContent)`
- How to handle a `<stencila-array>` or a `<stencila-thing> {Person}` inside a `Paragraph`

Screen Shot 2019-08-21 at 16 08 21

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,8 @@
   "devDependencies": {
     "@gerrit0/typedoc": "0.15.5",
     "@stencila/dev-config": "1.1.3",
+    "@testing-library/dom": "^6.1.0",
+    "@testing-library/jest-dom": "^4.0.1",
     "@types/async-lock": "1.1.1",
     "@types/content-type": "1.1.3",
     "@types/escape-html": "0.0.20",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "check:deps-used": "dependency-check --missing .",
     "check:deps-unused": "dependency-check --unused --no-dev --ignore-module @stencila/schema --ignore-module @stencila/thema .",
     "cli": "npx ts-node --files src/cli",
-    "build": "tsc && cp -r src/codecs/pandoc/templates/. dist/codecs/pandoc/templates/",
+    "build": "tsc -p tsconfig.prod.json && cp -r src/codecs/pandoc/templates/. dist/codecs/pandoc/templates/",
     "docs": "npm run docs:readme && npm run docs:dogfood && npm run docs:ts",
     "docs:readme": "markdown-toc -i --maxdepth=4 README.md",
     "docs:dogfood": "npx ts-node --files docs.ts",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "content-type": "^1.0.4",
     "datapackage": "^1.1.1",
     "escape-html": "^1.0.3",
-    "fp-ts": "^2.0.0",
+    "fp-ts": "^2.0.5",
     "fs-extra": "^8.1.0",
     "get-stdin": "^7.0.0",
     "github-slugger": "^1.2.1",
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,5 +2,13 @@
   "extends": "@stencila/eslint-config",
   "rules": {
     "@typescript-eslint/strict-boolean-expressions": 0
-  }
+  },
+  "overrides": [
+    {
+      "files": ["*.test.ts"],
+      "env": {
+        "jest": true
+      }
+    }
+  ]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -5715,9 +5715,9 @@
       "dev": true
     },
     "fp-ts": {
-      "version": "2.0.0",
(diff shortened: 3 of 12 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "check:deps-used": "dependency-check --missing .",
     "check:deps-unused": "dependency-check --unused --no-dev --ignore-module @stencila/schema --ignore-module @stencila/thema .",
     "cli": "npx ts-node --files src/cli",
-    "build": "tsc && cp -r src/codecs/pandoc/templates/. dist/codecs/pandoc/templates/",
+    "build": "tsc -p tsconfig.prod.json && cp -r src/codecs/pandoc/templates/. dist/codecs/pandoc/templates/",
```
nokome: Thank you for fixing this long standing annoyance :clap:

nokome:
Thanks for this @alex-ketch, looks good. 

I think we should start testing the generated HTML using the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/) to test HTML generated by Encoda. Do we need to have either JSON-LD or Microdata? What happens if we use both? etc. I'll create an issue for that.

Somewhat related to all this is a comment that I stumbled on yesterday: https://github.com/w3c/respec/issues/1503#issuecomment-366221786 an idea to mix JSON-LD in the `<head>` with HTML Microdata in the `<body>`.

With respect to `<stencila-thing>` and the other custom elements: given that we are moving to using Microdata, IMO we should not use these anymore. They were intended to provide additional semantics for content (a) for machine readability, (b) for progressive enhancement via web components. But using semantic elements e.g. `<time>` and Microdata is better for (a) and can still be used for (b) (by parsing the DOM for `itemtype` etc). 

Just to clarify what I mean by semantic HTML allowing for progressive enhancement. I imagined it would allow for things like "hovercards" when you however over some content, like Github does when I hover over some of the links above:

image

Github's HTML for that link is,

```html
<a href="https://github.com/stencila/thema/pull/10" data-hovercard-type="pull_request" data-hovercard-url="/stencila/thema/pull/10/hovercard">Thema PR</a>
```

`Articles` could have all sorts of "rich" semantic nodes (inline or block content) like `Person`, `Protein`, `Gene`, `Species` etc with custom hovercards. The difference being that with Encoda, instead of using `data-hovercard-type` etc we use more semantic markup and let the theme decide the presentation e.g.

```html
<a href="https://species-database.org/Pinus+radiata" itemtype="Species"><span itemprop="name">Pinus radiata</span></a>
```

PR #236, opened 2019-08-23
Update HTML Codec README with new info and examples

alex-ketch:
This is the first one of this type of README that I've done, though I did try to use the other READMEs for guidance, I'd appreciate feedback on:
- The appropriateness of the level of technicality/detail
- Relevancy to the target user of this tool
- Consistency of language/grammar
- anything else you'd like to leave feedback on :)

Please noteNote:
- I started with the `Encoding` section, rather than `Decoding` as this progression of information seemed clearer to me, and easier to annotate in the HTML example.
- I intentionally did not use `import` `export` tags on some of the code blocks as:
  - the HTML content is manually annotated with inline comments that get lost otherwise. I thought this was the clearest way to communicate to the reader what was happening.
  - The counterexamples were throwing errors due to being unable to be appropriately `coerced`


nokome:
This PR has gotten really stale I am afraid. I suspect much of the documentation has become out of date. Rather than it lingering any longer, I will close it, but will not delete the branch, in case we want to refer to it in the future.

PR #246, opened 2019-08-29
Remove statically coded Thema theme names close

### LINKED ISSUE #245: HTML: Matching user entered theme to those in Thema
From [here](https://github.com/stencila/thema/pull/10#issuecomment-525860933), use `theme.toLowerCase()` [here]( https://github.com/stencila/encoda/blob/8c3c9387da3c9f6863859268bffee08554726ef4/src/codecs/html/index.ts#L441).

Also, it would be good to warn the user (and just provide unstyled HTML, rather than erroring) if the entered theme does not match any of those in Thema (use a `fs.readdir` of Thema's `theme` directory)?

alex-ketch:
Closes #245 

~~⚠️ This PR is dependent on https://github.com/stencila/thema/pull/12 and will fail until that PR is merged, and Thema updated in this PR~~ Done ✅ 

This PR:
- Removes hard-coded theme names, and instead uses the `themes` object exposed by Thema
- Adds a `defaultEncodeOptions` object that is used as the fallback value for `encode` functions, this was done to better reflect the fact that a `theme` value is always present and while optional, it's never `undefined`. The `defaultEncodeOptions` should also allow for easier future additions down the road.

### DIFF
--- a/docs.ts
+++ b/docs.ts
@@ -4,9 +4,10 @@
  * to a tree of HTML in `docs`.
  */
 
+import * as logga from '@stencila/logga'
+import { themes } from '@stencila/thema'
 import { DirCodec } from './src/codecs/dir'
 import * as vfile from './src/util/vfile'
-import * as logga from '@stencila/logga'
 
 const { decode, encode } = new DirCodec()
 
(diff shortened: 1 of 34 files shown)


alex-ketch:
Thanks for the feedback @beneboy! I initially thought that since the class was an abstract one it wouldn't inherit, but that's not the case. I added a private property to the `Codec` class to make future code reorganization easier.

PR #247, opened 2019-08-29
Pandoc updates

nokome:
This adds handling for nodes that are currently not handled by the `pandoc` codec. The impetus for this wanting to demo conversion of a published article from JATS to Word and back again. This took a long time and resulted in a messy `.docx` because any unhandled node gets converted to a rPNG.

- [x] `Superscript` and `Subscript` - straightforward.
- [ ] `Cite` and `CiteGroup` - currently implemented but needs more consideration, in particular on how these get eventually encoded to `docx`
- [ ] `Math` - this is something that is often requested and seems well supported in Pandoc - but we do not yet have support in the schema for these node types

This is WIP and, given our current workplan, I'm just parking it here to come back to when conversion to/from Word is more of a priority.

### DIFF
--- a/src/codecs/pandoc/__file_snapshots__/cite.yaml
+++ b/src/codecs/pandoc/__file_snapshots__/cite.yaml
@@ -1,0 +1,46 @@
+type: Article
+title: Untitled
+authors: []
+bibliography:
+  - author:
+      - family:
+          type: Paragraph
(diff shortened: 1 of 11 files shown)


nokome:
Closing as these changes have been merged and math is dealt with in #383

PR #249, opened 2019-08-29
Codecs for http and journals

nokome:
This PR aims to:

- test the functioning of the `http` codec e.g. format resolution from media types and/or filename extension

- add codecs that fetch JATS from journals (currently `elife` and `plos`) - these were initially developed for generating test fixtures but having them as standalone codecs is also useful, especially for demos.


src/codecs/elife/elife.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  // Unlink to remove references to local files (which are non-deterministric)
+  const unlinked = await unlink(node)
+  // Convert to YAML
+  return await vfile.dump(await yaml.encode(unlinked))
+}
+
+
+jest.setTimeout(30 * 1000)
+
+test('sniff', async () => {
```
alex-ketch: A suggestion for future PR, but I'd recommend not grouping a lot of `expect` calls like this, especially when they're testing different bits of logic (like whitespace in some of the cases below).
Reason being that when a test fails, it's much harder to know why, and which expect call was the culprit.
Instead using a `describe` block, with multiple `test` cases named after the scenarios they're testing can be clearer and easier to debug.

nokome:
> Is the Pandoc Cite implementation something that remains to be done in this PR, or as a supplemental one?

Yes, that is being done in #247

Issue #250, opened 2019-08-30
MD: Unsupported node types


nokome:
Given current priority on the `md` codec, this is a "meta" issue to record some of the node types that are currently unsupported. These were found when attempting to convert real JATS journal articles to Markdown (see below; although that is not a common use case we obviously want to be able to support the reverse i.e. Markdown to JATS so should support these).

- [ ] Subscript & Superscript- use Pandoc's [syntax extensions](https://pandoc.org/MANUAL.html#superscripts-and-subscripts) and [remark-sub-super](https://github.com/zestedesavoir/zmarkdown/tree/master/packages/remark-sub-super)
- [ ] Math - requires adding `Math` and `MathBlock` (consistent with `Quote` and `QuoteBlock`) to schema; use [remark-math](https://github.com/remarkjs/remark-math)
- [ ] Cite - use Pandocs `@target-id` syntax extension; probably requires a custom remark plugin for decoding/encoding
- [ ] Figure - probably a generic block extension, Pandoc uses "div" wit "fig" class (?)
- [ ] Collection

Also worth noting, when encoding complete articles to Markdown, the YAML header becomes large because it get's filled with many references. The conventional approach is to instead encode the references as BibTex as a separate file e.g.
```yaml
title: My article
references: references.bib
```

```bash
./encoda convert https://elifesciences.org/articles/45562v3 temp.md

⚠ WARN  encoda:jats Using default decoding for JATS element name: "inline-formula"
⚠ WARN  encoda:jats Using default decoding for JATS element name: "inline-formula"
⚠ WARN  encoda:jats Using default decoding for JATS element name: "disp-formula"
⚠ WARN  encoda:jats Using default decoding for JATS element name: "label"
⚠ WARN  encoda:jats Using default decoding for JATS element name: "inline-formula"
⚠ WARN  encoda:md No Markdown encoder for Stencila node type "Cite"
⚠ WARN  encoda:md No Markdown encoder for Stencila node type "Figure"
⚠ WARN  encoda:md No Markdown encoder for Stencila node type "Math"
⚠ WARN  encoda:md No Markdown encoder for Stencila node type "Math"
⚠ WARN  encoda:md No Markdown encoder for Stencila node type "Superscript"
⚠ WARN  encoda:md No Markdown encoder for Stencila node type "Subscript"
```

PR #251, opened 2019-08-30
feat(Markdown): Add handling of Superscript and Subscript nodes

nokome:
Addresses one of the items listed in #250


src/codecs/md/index.ts
```diff
@@ -829,6 +838,48 @@ function encodeDelete(delet: stencila.Delete): MDAST.Delete {
...
+
+/**
+ * Encode a Stencila `Subscript` as a MDAST `text` node with surrounding tildes.
+ *
+ * This assumes that there is only `string`s in the `content` of the subscript.
+ */
+const encodeSubscript = (sub: stencila.Subscript): MDAST.Text => {
+  return {
+    type: 'text',
+    value: `~${sub.content.map(item => item !== null ? item.toString() : '').join()}~`
```
alex-ketch: Think it might be more resilient if we use `JSON.stringify(item)` here, as most non-string types will throw here.
Also, the documentation says that we assume there is only `string` in the `content`, so should we be filtering out non-`string` values instead?

nokome: Yes, that is a better approach: filter out, and warn of, non-`string` content

Issue #234, opened 2019-08-21
HTML: Test encoding of structured data


nokome:
Start using the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool) to test the machine readability of JSON-LD and Microdata that we encode into HTML. To do:

- [ ] Create some test fixtures of `Article`, `Datatable` etc (in JSON, to decouple from other codecs, but for convenience, perhaps generated from Markdown, JATS, CSV etc)
- [ ] Publish HTML generated for these fixtures to https://stencila.github.io/encoda/src/codecs/html/tests (add to docs build) so that they can be accessed from the GSDTT and we can put a link in the `html` codec's `README.md` like this one (which tests Encoda's own home page):

https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fstencila.github.io%2Fencoda

At the time of writing there is [no public API](https://support.google.com/webmasters/thread/2561708?hl=en) for the GSDTT. There is https://github.com/padosoft/laravel-google-structured-data-testing-tool but it is for PHP and uses an undocumented API. So, doing automated testing is likely to be more trouble than it is worth.

nokome:
Just to document one of the issues here. Example HTML generated by 

```bash
./encoda convert https://elifesciences.org/articles/45187v2 temp.html
```

produces many errors on GSDTT, including two for each `itemprop=citation`

```html
<li id="bib55" itemscope="true" itemtype="https://schema.org/CreativeWork"
itemprop="citation"><span itemprop="title" itemscope="true">Fusion stage of HIV-1 entry
    depends on Virus-Induced cell surface exposure of phosphatidylserine</span>
<ol itemprop="authors" itemscope="true">
    <li itemscope="true" itemtype="https://schema.org/Person" itemprop="author"><span
        itemprop="familyName">Zaitseva</span><span itemprop="givenName">E</span></li>
    <li itemscope="true" itemtype="https://schema.org/Person" itemprop="author"><span
        itemprop="familyName">Zaitsev</span><span itemprop="givenName">E</span></li>
    <li itemscope="true" itemtype="https://schema.org/Person" itemprop="author"><span
        itemprop="familyName">Melikov</span><span itemprop="givenName">K</span></li>
    <li itemscope="true" itemtype="https://schema.org/Person" itemprop="author"><span
        itemprop="familyName">Arakelyan</span><span itemprop="givenName">A</span></li>
[...]

PR #257, opened 2019-09-03
HTML: Improve Microdata

nokome:
Makes a number of changes based on testing against Google Structured Data Testing Tool see #234:

- remove JSON-LD  from head (currently not possible to use both JSON-LD and Microdata(?))
- replace invalid `itemprop`s (names that are not schema.org properties) with classes
- remove unnecessary use of `itemscope`
- add Microdata to top level `<article>`


nokome:
These changes result in GSDTT detecting an `Article` with `citations`. Adding more metadata e.g. `authors` still to be done.

image

snip

image

PR #268, opened 2019-09-11
HTML: Refactor decoding

nokome:
This refactors the approach to decoding HTML to handle the container elements such as `div` and `span` where we wish to return an array of nodes.


nokome:
Closing as now part of #269

PR #270, opened 2019-09-11
feat(HTML): Wrap CodeChunks in custom Web Component

alex-ketch:
Wraps `CodeChunk` nodes in necessary tags to work with the custom component from Designa

⚠️ Depends on https://github.com/stencila/designa/pull/16
Once that PR is merged and deployed on NPM/Unpkg, the [necessary script](https://github.com/stencila/encoda/pull/270/files#diff-c21a3c7e37c50b46ffa1ea7be006181bR545-R548) tag will hydrate and progressively enhance the custom components.
This also means that we can iterate and enhance existing components without having to re-encode/update documents. It is currently constrained to serve latest package below `v1`, so allows us to push breaking changes down the road if needed.

### DIFF
--- a/src/codecs/html/index.ts
+++ b/src/codecs/html/index.ts
@@ -25,15 +25,15 @@ import { html as beautifyHtml } from 'js-beautify'
 import jsdom from 'jsdom'
 import JSON5 from 'json5'
 import path from 'path'
+import { VFileContents } from 'vfile'
 import { columnIndexToName } from '../../codecs/xlsx'
+import { logWarnLossIfAny } from '../../log'
 import { compactObj, isDefined, reduceNonNullable } from '../../util'
 import bundle from '../../util/bundle'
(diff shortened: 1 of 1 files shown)


src/codecs/html/index.ts
```diff
@@ -539,7 +541,11 @@ function generateHtmlElement(
       }),
       h('meta', { 'http-equiv': 'X-UA-Compatible', content: 'ie=edge' }),
       themeCss,
-      themeJs
+      themeJs,
+      h('script', {
+        src: 'https://unpkg.com/@stencila/components@<=1',
+        type: 'text/javascript'
+      })
```
nokome: ~~This is using unversioned JS which makes sense for now but we should maybe think about pinning it later.~~ Hah, got pinned as I was writing that!

Also, I envisaged that the Web Component JS would integrate here, into Encoda, via Thema, rather than directly. Is that the long term plan?

PR #274, opened 2019-09-13
Wrap CodeOutput in <output> tags, Encode CodeExpressions inline HTML elements

alex-ketch:
Connects to https://github.com/stencila/designa/pull/17
Once the above PR is merged, we can add `@stencila/components` as a dependency and use the component exposed slot names in here for more resilient HTML generation.

- Wrap CodeOutput in <output> tags 
- Encode CodeExpressions as valid inline HTML elements

### DIFF
--- a/src/codecs/rnb/index.ts
+++ b/src/codecs/rnb/index.ts
@@ -138,7 +138,7 @@ export class RnbCodec extends Codec implements Codec {
       const html = container.innerHTML
       container.innerHTML =
         html.slice(0, begin) +
-        `<stencila-codeexpression>
+        `<stencila-codeexpression programming-language="r">
           <code class="language-r" slot="code">${source}</code>
           <span slot="output">${output}</span>
         </stencila-codeexpression>` +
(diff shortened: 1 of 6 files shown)


alex-ketch:
This should be ready for another look now, thanks! 🙇

src/codecs/rnb/index.ts
```diff
@@ -138,7 +138,7 @@ export class RnbCodec extends Codec implements Codec {
       const html = container.innerHTML
       container.innerHTML =
         html.slice(0, begin) +
-        `<stencila-codeexpression>
+        `<stencila-codeexpression programming-language="r">
```
nokome: To be consistent with https://github.com/stencila/designa/pull/17#discussion_r324853763 shall we change to `<stencila-code-expression>` and `<stencila-code-chunk>` as part of this PR?

alex-ketch: I was planning on changing things over in an upcoming PR, once I expose the `elementName` from the components

nokome: OK, thanks

PR #300, opened 2019-09-30
IPYNB: Decode notebook language

### LINKED ISSUE #290: IPYNB: CodeChunk programmingLanguage is not supported in decode
When decoding a Jupyter Notebook. The language of the notebook should be applied to `CodeChunk` nodes that are decoded from it.

nokome:
Extract the notebook language from it's metadata and use as `CodeChunk.programmingLanguage`. Closes #290


src/codecs/ipynb/ipynb.test.ts
```diff
@@ -27,10 +27,7 @@ describe('decode', () => {
     )
   })
 
-  test.skip('running-code', async () => {
-    // TODO: This test is currently failing, seemingly due to a bug
-    // in `md.stringifyHTML` which deals with HTML in HTML. Reinstate
-    // this when that is fixed.
+  test('running-code', async () => {
```
alex-ketch: 🎉

PR #305, opened 2019-10-09
fix: Decoding/Encoding CodeChunk outputs in MD

beneboy:
This fix seems to work with test and my MD->DocX now goes OK.


src/codecs/md/index.ts
```diff
@@ -627,13 +641,9 @@ function encodeCodeChunk(chunk: stencila.CodeChunk): Extension {
   }
   nodes.push(codeBlock)
 
-  // Separate the `output` with a `ThematicBreak`
   if (outputs && outputs.length) {
-    let index = 0
     for (const output of outputs) {
-      if (index !== 0) nodes.push({ type: 'ThematicBreak' })
```
beneboy: I removed this because it didn't seem to be necessary.

nokome: The intention of this was to use a `ThematicBreak` as a separator between the individual `outputs` of a `CodeChunk`. Without it, when decoding a `CodeChunk` you can't tell where one output ends and another begins.

src/codecs/md/index.ts
```diff
@@ -607,6 +608,19 @@ function decodeCodeChunk(ext: Extension): stencila.CodeChunk {
...
+
+        outputContainer.content.forEach(output => {
+          if (codeChunk.outputs === undefined) return // TS being dumb
+          if (typeof output === 'string') codeChunk.outputs.push(output.trim())
+          else codeChunk.outputs.push(output)
+        })
+      }
+    }
   }
   return codeChunk
```
nokome: While we are here, I suggest we refactor this to use factory functions:

```ts
const { programmingLanguage, meta, text = '' } = codeBlock
...
return stencila.codeChunk(text, { programmingLanguage, meta })
```

That compacts the object to exclude `undefined` properties, so then you don't need `if(programmingLanguage)` etc

nokome:
@beneboy IMO, this is good to be merged. Please go ahead if tests pass and you are happy with it.

PR #308, opened 2019-10-11
feat(PDF): Make reproducible using XMP metadata

nokome:
Despite the best efforts of many, PDF still live on strong in academic publishing. This PR applies Encoda's _"if you can't beat 'em, encode in 'em"_ approach to PDFs. It uses [XMP](https://en.wikipedia.org/wiki/Extensible_Metadata_Platform) to encode an entire node within a PDF. This make the round-trip conversion to PDF and back again completely lossless.

There is not much to do to finish this off to be production ready. But in the longer term, improvements that could be made to reduce the size of the PDF including not duplicating images in the JSON and the PDF (this would be possible by extracting the images from the PDF and having some sort of special reference in the JSON to them).


src/codecs/xml/README.md
```diff
@@ -0,0 +1,17 @@
+# XML codec
+
+The `xml` codec provides an alternative to the other document interchange formats such as `json` and `yaml`. It is used within Encode to embed a document within the XML metadata of PDF documents.
```
alex-ketch:
```suggestion
The `xml` codec provides an alternative to the other document interchange formats such as `json` and `yaml`. It is used within Encoda to embed a document within the XML metadata of PDF documents.
```

Typo?

src/codecs/xml/index.ts
```diff
@@ -0,0 +1,167 @@
...
+      return xml.text(elem)
+    case 'array':
+      return elements.map(decodeElem)
+    default:
+      const node = elements
+        .map(child => {
+          const key = xml.attr(child, 'key') || ''
+          return { [key]: decodeElem(child) }
+        })
+        .reduce((prev, curr) => ({ ...prev, ...curr }), {})
```
alex-ketch: Are the `keys unique within a collection of elements?
Also, instead of iterating over the collection twice we can combine the `map` operation within the `reduce`.
Something like:

```ts
.reduce((prev, curr) => {
    const key = xml.attr(curr, 'key') || ''
    return {
        ...prev,
        [key]: decodeElem(curr)
    }
}, {})
```

nokome: Yes, the `keys` are unique for this encoding - just as the keys of a JSON object are unique. There are no guarantees of course that some externally generated XML won't have duplicate keys - but I don't think we need to concern ourselves with that malformed content. Thanks for the suggested change, will make it.

PR #352, opened 2019-12-11
Expose as a Executa executor

nokome:
This implements the Executa executor API interface by extending `Listener`.

To test, do an `npm install` here (which will register Encoda on the machine), then in Executa do:

```bash
npm run cli -- serve --http --debug
```

Then hit the decode end point (here using `httpie`):

```bash
http POST :8000/decode content='# H1' format=md
```

There are a number of `TODO`s to address. Ping @alex-ketch

### DIFF
--- a/install.js
+++ b/install.js
@@ -39,4 +39,7 @@ async function load(modulePath) {
   // Install Pandoc binary
   const pandoc = await load('./dist/codecs/pandoc/binary')
   await pandoc.install()
+  // Register Encoda as an executor
+  const encoda = await load('./dist/encoda')
+  encoda.run('register')
 })()
--- a/install.js
+++ b/install.js
@@ -39,7 +39,4 @@ async function load(modulePath) {
   // Install Pandoc binary
   const pandoc = await load('./dist/codecs/pandoc/binary')
   await pandoc.install()
-  // Register Encoda as an executor
-  const encoda = await load('./dist/encoda')
-  encoda.run('register')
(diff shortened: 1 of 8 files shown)


install.js
```diff
@@ -39,4 +39,7 @@ async function load(modulePath) {
   // Install Pandoc binary
   const pandoc = await load('./dist/codecs/pandoc/binary')
   await pandoc.install()
+  // Register Encoda as an executor
+  const encoda = await load('./dist/encoda')
+  encoda.run('register')
```
alex-ketch: I'm probably missing context here, but I'm not sure if we should be running this as a post-install step.
I believe it'll be triggered when installed as a dependency (e.g. by schema), and I'm not sure what happens if it's registered multiple times.
Overall I think I'd rather see it as an explicit step that gets called upon desire to use it.

src/encoda.ts
```diff
@@ -0,0 +1,87 @@
+import {
+  Listener,
+  // TODO: Now that Executa re-exports logga
+  // and schema, shall we use those instead of
+  // potentially having version conflicts
```
alex-ketch: I think if we have a compatible version range defined, it leaves room removing `logger` or `schema` changes to Executa down the road. But I'm not set on this either way.

install.js
nokome: Yes, I think you are right. I was tempted by avoiding the extra step.

PR #376, opened 2020-01-14
feat(eLife): Automatically fetch the most recent version of article

### LINKED ISSUE #368: eLife: automatically fetch the most recent version
The [`elife` codec](https://github.com/stencila/encoda/blob/master/src/codecs/elife/index.ts) recognises a eLife article URL, parses out the article number and fetches the XML from the eLife Github repo. This is useful because it allows a user to copy and paste a eLife URL into the Hub and have it fetched automagically.

However, by default it will fetch the `v1` XML, not the most recent. This causes errors, because often the `v1` XML contains no body e.g

```bash
$ ./encoda convert https://elifesciences.org/articles/45562
⚠ WARN  encoda:elife Article 45562v1 has no body. You may wish to try another version.
⚠ WARN  encoda:jats Using default decoding for JATS element name: "article-title"
```

The user can explicitly add the version number to the URL e.g. `encoda convert https://elifesciences.org/articles/45562v3`. But that requires them to know the version.

It should be possible to query the Github API to find the most recent version for an article and fetch that.

nokome:
Closes #368.


nokome:
Over in https://github.com/stencila/encoda/issues/368#issuecomment-574095550, @giorgiosironi noted that the most recent version of an article is available at: `https://elifesciences.org/articles/{article}.xml` (and specific versions are available at `https://elifesciences.org/articles/${article}v${version}.xml`. Using these endpoints would simplify the code - in particularly negating the need for querying the Github API for the most recent version. I see two options:

1. delete the existing code that uses the Github API to get the latest version (but add a comment that permalinks to that code) and instead just fetch `https://elifesciences.org/articles/{article}.xml`

2. retain the existing code but have a switch, `const ELIFE_ARTICLE_SOURCE = 'elifesciences'`, which would allow us to fallback to using Github at a later time if for some reason `https://elifesciences.org/articles/{article}.xml` became unavailable.

I am leaning towards 1, since it results in less code (generally a good thing) and we can always reinstate the old code if necessary. @alex-ketch @beneboy Thought?

alex-ketch:
I also lean towards the first option for the same reasons @nokome

PR #392, opened 2020-01-22
Linting for strict boolean expressions

### LINKED ISSUE #199: Approach to dealing with non-strict-boolean-expressions
In `.eslintrc` we turn off `strict-boolean-expressions`:

```json
{
  "extends": "@stencila/eslint-config",
  "rules": {
    "@typescript-eslint/strict-boolean-expressions": 0
  }
}
```

If you renable it, you get 447 errors like:

```
/home/nokome/stencila/source/encoda/src/util/xml.ts
   24:6   error  Unexpected non-boolean in conditional  @typescript-eslint/strict-boolean-expressions
```

We've had a discussion in https://github.com/stencila/dev-config/issues/4 regarding whether check is too strict or not. We've also come across a bug which would have been avoided if this check was enabled: #197 

Many (most?) of these `Unexpected non-boolean in conditional` errors are in places like this when decoding:

https://github.com/stencila/encoda/blob/b85d598b54523829425ea89b4189571d29e271fc/src/codecs/ipynb/index.ts#L119-L121

which could be refactored to

```ts
const { title = 'Untitled', authors = [], orig_nbformat, ...rest } = notebook.metadata
```

or 

https://github.com/stencila/encoda/blob/6966088f5036b93ce1ec143b10187b71650be434/src/codecs/jats/index.ts#L265-L275

which with compacting factory functions could now be refactored to:

```ts
const surname = child(name, 'surname')
const familyNames = surname ? text(surname).split(/\s+/) : undefined
//... etc
return stencila.person({
  givenNames,
  familyNames,
  honorificPrefix,
  honorificSuffix
})
```

So, I propose to:

- renable `strict-boolean-expressions` at the package level
- add `/* eslint-disable @typescript-eslint/strict-boolean-expressions */` to every file that currently needs it
- as we tidy up and refactor each codec remove dangerous expressions and eventually remove these disabling comment

nokome:
Closes #199.


src/codecs/dir/index.ts
```diff
@@ -2,6 +2,16 @@
...
+/**
+ * Hello contributor 👋! If you are working on this file, please
+ * endeavor to remove the need for the following `eslint-disable` line 🙏.
+ * Remove the line and run `npx eslint path/to/this/file.ts` to
+ * see which code needs some linting ❤️.
+ * See https://github.com/stencila/encoda/issues/199 for suggestions
+ * on how to refactor code to avoid non-strict boolean expressions.
+ */
+/* eslint-disable @typescript-eslint/strict-boolean-expressions */
+
```
alex-ketch: 💖

PR #393, opened 2020-01-22
fix(MD): Don't coerce decoded MD AST, avoiding false null values

### LINKED ISSUE #307: Commented out blocks are rendered as 'null'
As can be seen in the screenshot below, when a commented out section is encountered in a Markdown file, it gets encoded as `null`. I believe the expected behaviour should be to omit the section, or include it as invisible comment if the target format supports it.

Screen Shot 2019-10-09 at 23 04 56

The source MD is included below:

```md
### Footnotes & References

<!-- Footnote content, such as this[^This is an example of a footnote.], will be rendered as… -->

*TODO*
```

alex-ketch:
As discussed over video, this removes the `coerce` call on the decoded MD schema, closes #307 
Down the road we might look into reinstating coercion in a single top-level place in the `index.ts` file.

### DIFF
--- a/src/codecs/md/index.ts
+++ b/src/codecs/md/index.ts
@@ -5,11 +5,11 @@
 import { getLogger } from '@stencila/logga'
 import stencila, {
   isBlockContent,
+  isInlineContent,
   isListItem,
   nodeIs,
   nodeType,
-  TypeMapGeneric,
-  isInlineContent
+  TypeMapGeneric
 } from '@stencila/schema'
 import * as yaml from 'js-yaml'
 import JSON5 from 'json5'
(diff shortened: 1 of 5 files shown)


src/codecs/bib/__file_snapshots__/article.bib
```diff
@@ -1,9 +1,9 @@
-@article{Smith2010title,
+@article{Smith2009title,
 	title = {The title of the article},
 	author = {Smith, Jane and Jones, Jim},
-	date = {2010-01-01},
-	year = {2010},
-	month = {1},
-	day = {1},
+	date = {2009-12-31},
```
nokome: This is related to timezone differences between machines. I thought I had resolved it in the `csl` codec (or maybe the `date` codec) but obviously not. Could you please (a) revert this change so that tests pass on master when we merge this (b) raise an issue about this.

Issue #389, opened 2020-01-21
Template codec


nokome:
Several use cases have come up where it would be useful to have some sort of templating. This is especially true when HTML is being generated e.g. having a template for the page layout and Encoda populating the `<main>` tag with converted content.

Rather than provide a templating solution for HTML only, this proposes to add a special `template` codec that can be used to with all other codecs. It would use a general purpose templating engine like Nunjucks and should ideally allow the user to specify templates not just for the top level node e.g. `Article`, but also for nested descendant nodes so the default encoding for a codec can be overridden on a node type basis. 

For example, using a single top level template:

```
encoda convert article.json article.html --template article.njk
```

Using a folder of templates to override some the encoding certain node types:

```
templates/
    Heading.njk
    Link.njk
```

```
encoda convert article.json article.html --template templates/
```

Some exploration needs to be done on how this mix of default / custom encoding could be implemented but with [custom filters](https://mozilla.github.io/nunjucks/api.html#custom-filters) should be able to achieve good DX.

This seems like a high priority given current explorations by others on how to use / integrate Encoda.

nokome:
Closing in favor of allowing template languages within `CodeChunk`s ie. ability to generate content within a document based on document variables.

PR #394, opened 2020-01-23
feat(Templates): Add templating codecs

nokome:
See #389. This is a spike to explore how to add templating functionality. I am happy with the approach taken here (piggybacking on top of existing templating engines) but also think we should consider an additional "Typescript/Javascript template codec", similar to that [`11ty.js`](https://www.11ty.dev/docs/languages/javascript/) and as briefly discussed with @alex-ketch.

Note how the custom `dump` filter in the `njk` codec allows using the default codecs, including mixing formats.

Before merging:

- get to work with the `dir` codec so that can render a whole directory of files using a template.
- more docs
- docs on security implications
   - rendering untrusted templates e.g. https://mozilla.github.io/nunjucks/templating.html
   - serving rendered output `<script>` etc tags created by a user template


nokome:
This PR has languished for over a year now. Whilst potentially useful there obviously isn't a lot of demand for this sort of template functionality (yet). So closing.

PR #399, opened 2020-01-27
Add microdata attributes

alex-ketch:
This PR adds MicroData `itemprop` attributes to most HTML elements, which we are using as semantic element selectors for Thema. This work relates to https://github.com/stencila/thema/issues/25.


nokome:
@alex-ketch I've made 4 commits that are hopefully fairly self explanatory. The last one is the most important - it uses the JSON-LD `@context` (which is itself based on the `@id`s defined in the schema `*.yaml` files) to map between type and property names and `itemtype` and `itemprop`). That should be more accurate and need less maintenance than a manually written `switch`. Also it introduces the `encodeAttrs` function which avoids a bunch of repetitive boilerplate code.

I will continue tomorrow focusing on getting more consistency in encoding of both `itemtype` and `itemprop` across node types and creating a function `encodeArrayProperty` to deal with the common pattern `h( 'ol', { attrs: { [stencilaItemProp]: ...`.

I have temporarily `skip`ed one encoding related test but will reinstate tomorrow.

I think it would be useful to add #234 to this PR, perhaps you could look into making use of  https://github.com/glitchdigital/structured-data-testing-tool?

src/codecs/html/html.test.ts
```diff
@@ -117,17 +136,23 @@ describe('Encode & Decode cite nodes', () => {
...
+  const codeExpressionNode = codeExpression({
     text: 'x * 2',
     output: '42',
     programmingLanguage: 'python'
   })
-  const codeExpressionHTML =
-    '<stencila-code-expression programming-language="python"><code slot="text">x * 2</code><output slot="output">42</output></stencila-code-expression>'
+  const codeExpressionHTML = `
+    <stencila-code-expression
+      itemtype="${encodeMicrodataItemtype('CodeExpression')}"
```
alex-ketch: _Posting this ahead of full review:_
I would like us to move away from using string values as much as possible (here and elsewhere), and instead use Enums or simple object key/value pairs to make any updates easier and prevent mistakes.

nokome: Agreed. In the case of `encodeMicrodataItemtype` the parameter is `keyof Types` so Typescript will indeed catch errors:

```
    src/codecs/html/html.test.ts:147:43 - error TS2345: Argument of type '"CodeExpressio"' is not assignable to parameter of type '"Article" | "Cite" | "CiteGroup" | "Code" | "CodeBlock" | "CodeChunk" | "CodeExpression" | "Collection" | "CreativeWork" | "Datatable" | "Organization" | "Periodical" | "PublicationIssue" | ... 69 more ... | "VariableTypes"'.

    147       itemtype="${encodeMicrodataItemtype('CodeExpressio')}"
```

For `encodeMicrodataProperty` the parameter is a currently a string since we don't yet have an enumeration of all the properties across all types (but presumably we could make that?)

nokome:
Update:

- reorganised fixtures for better reuse
- used a fixture to test HTML encoding
- made a bunch of fixes based on copy and pasting [generated HTML](https://github.com/stencila/encoda/pull/399/files#diff-fbdaf2887f3eb830efa26c5e2c95722b) into Google SDTT
- decided that we really need to automate the above so added SDTT package

Next steps:

- get the `microdata.test.ts` to pass by encoding the things that Google wants
- the code feels very brittle - make it more robust **once** the Microdata data tests are part of CI 9so things don't break during refactoring!)

alex-ketch:
I was re-generating examples for Thema development and encountered the following error due to trying to encode `String` schema type as MicroData.

Screen Shot 2020-01-30 at 13 23 12

The relevant lines are here, but we don't have a `String` schema. Need to dig a bit more to get a better understanding of the recent changes, but should we try to use add the Schema.org `Text` instead?

https://github.com/stencila/encoda/blob/cd377d6ca2b77e7ded6aeefab53c752676f22a24/src/codecs/html/microdata.ts#L145-L157

nokome:
>  but should we try to use add the Schema.org Text instead?

Yes, I think we should - it is the only (?) `isPrimitve` type that is without  a mapping. Feel free to go ahead and do that. It will also need a `schema:Text` entry here: https://github.com/stencila/encoda/blob/cd377d6ca2b77e7ded6aeefab53c752676f22a24/src/codecs/html/microdata.ts#L117

nokome:
@alex-ketch I've rebased on `master` and pushed my recent work on fixing the microdata tests. I think the main thing to do is to ensure everything has `itemtype` for Thema to target. I don't think we need to worry further about `itemprop` at this time - as I mentioned the code can be improved wrt to that, but it's good enough for now and it would be good to merge this ASAP to progress Thema

nokome:
Oh, and also the two Microdata related `TODO:`s in the `html/index.ts` file :)

alex-ketch:
@nokome I've added further refinements to the encodings, and addressed TODOs inside the html file. We now use https://placeholder.com images as fallbacks for publisher logos and article hero images. If the fallbacks are used, they are inserted inside `meta` tags to avoid distracting from the content.

Going to merge this with the note that there are several schema nodes which were not being encoded to HTML, and are naturally missing MicroData attributes. They can be found commented out in the MicroData encoding test in `html.test.ts` file.

Issue #418, opened 2020-02-04
JATS: Decoding sections in backmatter


FAtherden-eLife:
eLife content contains sections within the backmatter.

These sections appear to not be decoded.

### Steps to reproduce
Visit [here](https://elifesciences.org/articles/52355/figures#files) and note the 'Additional files' section containing numerous downloadable assets.

In jats this is captured contained in a `<sec sec-type="supplementary-material">`.
The command
```
encoda convert https://elifesciences.org/articles/52355 52355.jsonld
```

Produces a jsonld file where this section is not present.

Issue #410, opened 2020-01-30
JATS: encoding/decoding of Acknowledgements (in <back>)


FAtherden-eLife:
Some eLife content has Acknowledgement sections, captured in jats in a `ack` element.

This does not seem to be decoded/encoded. For example the acknowledgements section [here](https://elifesciences.org/articles/42512#info) is not present in this output:
```
encoda convert https://elifesciences.org/articles/42512 42512.jsonld
```

Issue #409, opened 2020-01-30
JATS: Data availability section (in <back>)


FAtherden-eLife:
eLife content contains a Data availability section, which is captured in a `back/sec[@sec-type="data-availability"]` element.

This section is not encoded/decoded. For example, none of the content [here](https://elifesciences.org/articles/42512/figures#data) is in this file
```
encoda convert https://elifesciences.org/articles/42512 42512.jsonld
```

Issue #400, opened 2020-01-28
JATS: Decoding (and encoding) of Appendices


fred-atherden:
eLife content contains appendices, which can contain the same types of content which is allowed in the main body of an article - [example 1](https://elifesciences.org/articles/46112#appendix-1), [example 2](https://elifesciences.org/articles/46351#appendix-1).

Appendices could contain figures which an author might also want to be reproducible, executable etc.

Currently it looks like appendices aren't decoded. 
For example 
```
encoda convert https://elifesciences.org/articles/46112v4 elife-46112.jsonld
```
produces a JSON-LD file without that content present. 

(I'm unsure whether this is a decoding or encoding issue)

eLife consider this high priority.

PR #421, opened 2020-02-04
JATS: decoding back matter elements

nokome:
This is a draft PR that aims to address several issues related to decoding of JATS back matter: #400 #409 #410 #418


nokome:
Closing as stale (and pointless in current form :)

PR #425, opened 2020-02-07
test(Docker): Add testing within a Docker container

nokome:
The overall aim of this PR is to have a tested, reference `Dockerfile` for those wanting to run Encoda in a container (e.g. as in the [`stencila/executa` image](https://github.com/stencila/executa/blob/2db7f2425d47d39dd8f24569863222c421634b18/Dockerfile#L42))

This  makes substantial changes to the Dockerfile to:
 - support running Puppeteer inside a container and,
 - to fix the installation of this package

It also adds Docker testing to CI.

@giorgiosironi I think you mentioned that you had made some fixes to the Dockerfile. If you could review this PR, and add any further improvements, that would be appreciated.


Dockerfile
```diff
@@ -1,26 +1,48 @@
...
+# unsafe `--no-sandbox` option.
+RUN useradd --create-home encoda
+WORKDIR /home/encoda
+USER encoda
+
+# Copy over everything and install.
+# Do not do the dusual Node package / Docker optimization of copying over
+# `package.json` and installing in a seperate layer because that is not possible given
+# how our `install.js` script works.
+COPY --chown=encoda:encoda . .
```
giorgiosironi: the separate `chown -R` was a performance killer, so this helps.
Though is really any file changes in the root a trigger for a new `npm install`? For example, even a change to `README.md` (doesn't seem `.dockerignore`d) or a new line at the end of the `Dockerfile` (can't be `.dockerignore`d) would result in a new execution. Could list out explicitly the folders with various `COPY` instructions?

nokome: Yes, good point. I have in the past been bitten by not including all necessary files (IIRC) so have formed a habit of `COPY . .`, which is indeed sub-optimal. Well revise. Thanks!

PR #488, opened 2020-03-15
DOCX: Improve encoding

nokome:
This PR aims to improve encoding of `Article`s to DOCX. This includes node types that are currently not encoded (i.e. are lost):

- [ ] authors and affiliations
- [ ] date published (included as part of document properties, but not displayed)
- [x] figures
- [x] references `isPartOf` types `PublicationIssue`, `Periodical` and `pageStart` etc (these need to be converted in CSL codec)

Other improvements to be made include:

- [x] table and figure captions
- [ ] adding option to encode `CodeChunk` and `CodeExpression` nodes as an rPNG _with_ alttext etc, so that they can be represented in Google Docs.


nokome:
Although not all the check boxes have been checked, I'd like to merge this into `master` and create issues for those remaining issues (so that I can move on to work on the Hub).

src/codecs/types.ts
```diff
@@ -202,44 +203,87 @@ export abstract class Codec<
...
+  }
+
+  /**
+   * Transform a node after reading it.
+   *
+   * Makes any references to local files absolute.
+   * Derived classes may override this method to perform alternative
+   * transformations to the node, prior to writing it.
+   *
+   * @param filePath The path of the file
```
alex-ketch:
```suggestion
   * @param node The `stencila.Node` to transform
   * @param filePath The path of the file
```

src/codecs/types.ts
```diff
@@ -202,44 +203,87 @@ export abstract class Codec<
...
+   * Transform a node prior to writing it
+   *
+   * Depending on the `isBundle` encoding option,
+   * either converts media to data URIs, or copies them
+   * to a sibling folder.
+   * Derived classes may override this method to perform alternative
+   * transformations to the node, prior to writing it.
+   *
+   *
+   * @param node The `stencila.Node` to write
```
alex-ketch:
```suggestion
   * transformations to the node, prior to writing it.
   *
   * @param node The `stencila.Node` to write
```

src/util/content/ensureInlineContentArray.ts
```diff
@@ -6,13 +6,16 @@ import { TxtCodec } from '../../codecs/txt'
...
-      ? [...prev, node]
-      : 'content' in node
-      ? [...prev, ...node.content]
-      : [...prev, TxtCodec.stringify(node)]
-  }, [])
+  return (Array.isArray(nodes) ? nodes : [nodes]).reduce(
+    (prev: schema.InlineContent[], node) => {
+      return schema.isInlineContent(node)
+        ? [...prev, node]
+        : 'content' in node
```
alex-ketch: I don't think this check is enough. It'll be a false positive for things like [`QuoteBlock`s](https://github.com/stencila/schema/blob/e88891e22b1028e5c6610c5a60d18a298fb604e0/schema/QuoteBlock.schema.yaml#L17-L22) which have `BlockContent` arrays for `content`.

src/util/transform.ts
```diff
@@ -21,30 +21,42 @@ import * as stencila from '@stencila/schema'
  *
  * @param node The node to transform
  * @param transformer The transforming function. Should return the transformed node.
+ * @param resurse If a node is transformed, should transformation continue in to it's children?
```
alex-ketch:
```suggestion
 * @param recurse If a node is transformed, should transformation be performed on its children?
```

src/util/content/ensureInlineContentArray.ts
nokome: Good catch!. Fixed and added tests.

PR #600, opened 2020-06-19
feat(MD): Encode references in a separate file

### LINKED ISSUE #589: feat(Markdown): Encode references in a separate file
Currently, when encoding to Markdown based formats (i.e including RMarkdown), all article properties apart from `content` are placed in the YAML header. This includes the `references` property, which can be very long. For example doing:

```sh
encoda convert https://elifesciences.org/articles/52258 pastoll-et-al-2020.Rmd
```
results in a Rmd file with almost 3000 lines of YAML encoded references at the top. This is inconvenient (you have to scroll a long way down to get to the main content) and somewhat confusing for some users (at first it looks like a YAML document not a Rmd one).

The usual convention for a large set of references is to link to an external file, usually in BibTeX (see  for example, https://pandoc.org/MANUAL.html#option--bibliography). It should be fairly straightforward to implement encoding Rmd references into a separate YAML file, or even a separate BibTeX file given that we already have a BibTeX codec. Might be best to only use a separate file if there are more than, say 5, references.

/cc @mattnolan001 @FAtherden-eLife

alex-ketch:
Close #589

This PR implements a feature to extract references into a `<filename>.references.bib` file next to the Markdown file.
This is only done when the article contains more than 5 references.
Converting back from MD attempts to read in this references file again.

### DIFF
--- a/src/codecs/csl/index.ts
+++ b/src/codecs/csl/index.ts
@@ -38,7 +38,7 @@ export class CSLCodec extends Codec implements Codec {
 
     const content: string = await vfile.dump(file)
 
-    let csls
+    let csls: Csl.Data[]
     try {
       csls = await Cite.inputAsync(content, { forceType: format })
     } catch (error) {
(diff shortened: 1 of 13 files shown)


src/codecs/csl/index.ts
```diff
@@ -186,9 +204,11 @@ export const encodeCreativeWork = (
   logWarnLossIfAny('csl', 'encode', work, lost)
 
   const date = datePublished ?? dateModified ?? dateCreated ?? undefined
+
   return {
     type,
-    id: id !== undefined ? id : crypto.randomBytes(16).toString('hex'),
+    id,
+    'citation-label': id,
```
alex-ketch, line 211: The `citation-label` was added in order to preserve the `reference` ids used in the Markdown files.

PR #610, opened 2020-07-02
fix(MD): Store extracted references to a `bibliography` key

### LINKED ISSUE #609: RMarkdown: use bibliography not references
Currently, when encoding to RMarkdown the property `references` is used in the YAML header to refer to an external file of references. According to the Pandoc docs, https://pandoc.org/MANUAL.html#citations, this is valid for in-YAML references. However, when using Bookdown / Knitr a user has reported it was necessary to change this to `bibliography` e.g

```md
---
blbliography: some-file.bibtex
---
```

alex-ketch:
Close #609

### DIFF
--- a/src/codecs/md/index.ts
+++ b/src/codecs/md/index.ts
@@ -197,7 +197,9 @@ export function stringToMdast(content: string): UNIST.Node {
 type Frontmatter = Record<string | number, unknown>
 
 interface DecodeContext {
-  frontmatter: Partial<Exclude<stencila.CreativeWork, 'content'>>
+  frontmatter: Partial<
+    Exclude<stencila.CreativeWork, 'content'> & { bibliography: string }
+  >
 }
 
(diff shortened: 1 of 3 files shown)


src/codecs/md/index.ts
```diff
@@ -242,14 +244,16 @@ async function decodeRootArticle(
...
-
   const context: DecodeContext = {
     frontmatter: {
       ...loadedFrontmatter,
       references: await loadExtractedReferences(loadedFrontmatter),
     },
   }
 
+  // Remove the reference to the bibliography file as it has been inlined to the `references` key by this point
+  delete context.frontmatter.bibliography
```
alex-ketch, line 255: What do you think about this @nokome? Should we keep the reference to the file path of the `bibliography` file, or omit it from the `meta/frontmatter` as I've done here?

Issue #649, opened 2020-08-20
JATS: decoding table missing rowspan/colspan


asisiuc:
Jats lack rowspan and colspan attributes when converting to format JSON:

### Steps to reproduce

This is captured in the jats as:
```
  <thead>
    <tr>
      <th rowspan="2" align="left"></th>
      <th colspan="4" align="left">Uruguay</th>
      <th colspan="4" align="left">Latin american average</th>
    </tr>
  </thead>
```
After running
```
encoda convert -- convert ./../encoda/article.xml article.json --from jats
```
it is output as:
```
        {
          "type": "TableRow",
          "cells": [
            {
              "type": "TableCell",
              "content": []
            },
            {
              "type": "TableCell",
              "content": [
                "Uruguay"
              ]
            },
            {
              "type": "TableCell",
              "content": [
                "Latin american average"
              ]
            }
          ],
          "rowType": "header"
        },
```
In the converted result missing attributes colspan and rowspan.

PR #650, opened 2020-08-20
fix(JATS): Decoding table rowspan and colspan

asisiuc:
This is a PR to related to decoding of JATS table attributes rowspan and colspan: #649


alex-ketch:
Hi @asisiuc, thanks for the PR! I'll take a closer look at the changes soon.
Looking at the CI failure it seems that the failures are due to the expected changed in the snapshots (adding the `col/rowspan values) and need to be updated by running `npm run test -- -u`.
Could you run that command locally and commit the updated snapshots please?

asisiuc:
> Hi @asisiuc, thanks for the PR! I'll take a closer look at the changes soon.
> Looking at the CI failure it seems that the failures are due to the expected changed in the snapshots (adding the `col/rowspan values`) and need to be updated by running `npm run test -- -u`.
> Could you run that command locally and commit the updated snapshots please?

Hi, @alex-ketch I updated the snapshots. Thanks for your help!

nokome (changes requested):
Thanks for the PR @asisiuc, great to see this done.

Could you make a couple of relatively minor changes before we merge it:

- remove `src/__tests__/issues/74-write-epipe.html` (this file is generated during a test and is supposed to be cleaned up automatically but that obviously fails on your machine when you ran the tests)
- revert the change to `src/codecs/md/__file_snapshots__/mdReferences.md` (it seems that you are on Windows and that has changed the direction of the flashes; ideally this would be OS agnostic, but that is a fix for another time)
- only keep `colspan` and `rowspan` if they are not equal to one (this would avoiding adding lots unnecessary `colspan` and `rowspan` properties e.g in `src/codecs/jats/__file_snapshots__/plosone-0091296.yaml`)

Issue #651, opened 2020-08-21
JATS: decoding fpage and lpage


asisiuc:
Jats lack fpage and lpage tag when converting to format JSON:

### Steps to reproduce

This is captured in the jats as:
```
<fpage>1</fpage>
<lpage>40</lpage>
```
After running
```
encoda convert -- convert ./../encoda/article.xml article.json --from jats
```

In the converted result missing tags fpage and lpage for article.

PR #652, opened 2020-08-24
fix(JATS): Decoding for article fpage and lpage

asisiuc:
This is a PR to related to decoding of JATS for article convert tag fpage and lpage: #651


nokome:
Hi @asisiuc. Thanks for this. 

It would be great if you could add a JATS XML file with `<fpage>` and `<lpage>` in `<front>` in to `src/codecs/jats/__fixtures__`. None on the existing fixtures have that so there is no test coverage for these changes. Maybe you could drop in an example JATS XML file and add it to the list here:
https://github.com/stencila/encoda/blob/17cbff033064066894147007e1213873e72b1e7d/src/codecs/jats/jats.test.ts#L74-L85

and then run `npx jest -u`

Issue #653, opened 2020-08-24
JATS: decoding for article issue number with tag <issue>


asisiuc:
Jats lack issue tag when converting to format JSON:

### Steps to reproduce

This is captured in the jats as:
```
<issue>100</issue>
```
After running
```
encoda convert -- convert ./../encoda/article.xml article.json --from jats
```

In the converted result missing tag issue for article.

PR #654, opened 2020-08-25
fix(JATS): Decoding for article issue number

asisiuc:
This is a PR to related to decoding of JATS for article convert tag issue : #653
Require to have types in [https://github.com/stencila/schema/issues/207](https://github.com/stencila/schema/issues/207)

### DIFF
--- a/src/codecs/jats/__file_snapshots__/plosone-0091296.yaml
+++ b/src/codecs/jats/__file_snapshots__/plosone-0091296.yaml
@@ -124,6 +124,7 @@ isPartOf:
       name: Public Library of Science
     title: PLoS ONE
   volumeNumber: '9'
+issue: 3
 licenses:
   - type: CreativeWork
     content:
--- a/src/codecs/jats/__file_snapshots__/plosone-0091296.yaml
+++ b/src/codecs/jats/__file_snapshots__/plosone-0091296.yaml
@@ -101,30 +101,32 @@ identifiers:
     propertyID: 'https://registry.identifiers.org/registry/elocation-id'
     value: e91296
 isPartOf:
-  type: PublicationVolume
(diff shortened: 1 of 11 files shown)


nokome:
Thanks for this @asisiuc, all your PRs are appreciated!

I don't thing that we should add an `issue` property to the schema because it is already represented by the `issueNumber` property of the [`PublicationIssue`](https://schema.stenci.la/publicationissue) type. Instead, when decoding JATS `<article-meta><front>` you should take the same approach as we do in `decodeReferences` and extract `<volume>`, `<issue>` into the `isPartOf` property of the `Article`:

https://github.com/stencila/encoda/blob/40fd470042c2ed1866a892ff2ad2e4dec33c142d/src/codecs/jats/index.ts#L994-L1000

The difference for `<front>` is there is no `<source>` to provide the `Periodical.name`. Instead, it would be great if you could write a `decodeJournalMeta` function to create a [`Periodical`](https://schema.stenci.la/periodical) node from the JATS `<journal-meta>` element and use that.

The resulting JSON will be nested (this is actually YAML from one of the file snapshots):

```yaml
    isPartOf:
      type: PublicationIssue
      isPartOf:
        type: PublicationVolume
        isPartOf:
          type: Periodical
          name: PLoS Biol
        volumeNumber: 9
      issueNumber: 4
```

This is a little less convenient to encode to HTML but is consistent with schema.org.

asisiuc:
Hi @nokome , thanks for the message it was useful, after analyzing all the examples of articles we have in ```__fixtures__``` I came to the conclusion that in our case and for the examples from "plosone" the ```<issue>``` tag still comes directly in ```<article-meta>```, so we insisted on being the same idea as the reference.

Issue #668, opened 2020-09-03
unable to convert .docx to .ipynb


kayarre:
Unsure how to debug the error:
```
 ERROR encoda Cannot read property 'content' of undefined
```
I am trying to convert this available resource: https://github.com/NCAR/CfRadial/tree/master/docs
e.g.
./encoda convert ~/Downloads/CfRadialDoc-v2.1-20190901.docx ~/Downloads/CfRadialDoc-v2.1-20190901.ipynb

[2020-09-03T16_51_43_440Z-debug.log](https://github.com/stencila/encoda/files/5170148/2020-09-03T16_51_43_440Z-debug.log)

nokome:
Thanks for the detailed bug report @kayarre. I can confirm that I get the same error on Ubuntu 16.04 (:scream: yes, I need to get around to upgrading!). It appears to be an issue decoding this `.docx` because attempting to convert `.json` gives the same error.

I'm looking into it now.

PR #669, opened 2020-09-03
DOCX to MD and IPYNB

nokome:
This PR principally aims to fix the fatal bag reported in #668. However, it also makes other improvements in the conversion of a `.docx` file to `.md` (and by extension to `.ipynb`, the target format of the original bug report):

- [x] encoding of block content within table cells to Markdown (which only allows for inline content in table cells)
- [x] resolving of embedded image files
- [x] capturing table caption
- [ ] capturing table widths


nokome:
This seems to be in pretty good shape now (certainly much better than it was). 

@kayarre here is the Markdown file generated from the DOCX (it is part of the test I added):

https://github.com/stencila/encoda/blob/bb512bc6ab6ce315baedabdda286166f49960935/src/__tests__/issues/__file_snapshots__/668-CfRadialDoc-v2.1-20190901.md

And a Jupyter Notebook I generated from it: [668.ipynb.txt](https://github.com/stencila/encoda/files/5171997/668.ipynb.txt)

Let me know if there are obvious things that need to be fixed.

One thing to note is that the `md` codec currently encodes a `Subscript` node using surrounding tilde's e.g. `X~a~` but some Markdown parsers use that for ~strikethrough~ (see Github does; we use double tilde's `~~` for that). I'll create a separate issue for that.

Issue #667, opened 2020-09-03
JATS: Decoding tag <table-wrap-foot> missing in <table-wrap>


asisiuc:
Jats lacks ```<table-wrap-foot>``` tag content when converting to format JSON:

### Steps to reproduce

This is captured in the jats as:
```
<table-wrap id="table1" position="float">
  <label>Table 1.</label>
  <caption></caption>
  <table></table>
  <table-wrap-foot>
    <fn>
      <p>
        <italic>Source:</italic>
        CEPAL (2016) and World Bank (2016).
      </p>
    </fn>
    <fn id="T1_FN1" fn-type="general">
      <p><italic>Notes</italic>: a/ 2005&#x2013;2010; b/ 2010; c/ Percentage of population; d/ Population ages
        25&#x2013;59; e/ Population ages 15&#x2013;64; f/ Percentage of households.
      </p>
    </fn>
  </table-wrap-foot>
</table-wrap>
```
After running
```
encoda convert -- convert ./../encoda/article.xml article.json --from jats
```
In the converted result ```<table-wrap-foot>``` tag is missing content.

nokome:
Thanks for this @asisiuc .

For reference, the [JATS standard](https://jats.nlm.nih.gov/archiving/tag-library/1.1/element/table-wrap-foot.html) says that `<table-wrap-foot>` is a "Container element to hold the footnotes, general notes, or other paragraphs at the end of a table."

The main question is which property of a `Table` node should these notes get added to. I see two options:

1. append it the existing `caption` property; downside is that this would be "lossy" in that it lumps `<caption>` and `<table-wrap-foot>` and so wouldn't allow the two to be displayed separately e.g. in HTML, upside is that it's the easiest to implement

2. use it for one of the existing properties of a table https://schema.stenci.la/table: `content` or `description` (of the two, I suggest `description` since that is an schema.org property, whereas `content` is a custom extension of the Stencila schema.

If we were to go ahead with using description, it should be fairly straightforward to implement in `decodeTableWrap`, along the lines of how we handle `caption` e.g.

```ts
  const descriptionElem = child(elem, 'caption')
  const description =
    descriptionElem !== null && Array.isArray(descriptionElem.elements)
      ? decodeElements(descriptionElem.elements, state)
      : undefined

...

  return [
    stencila.table({
      id: decodeInternalId(attr(elem, 'id')),
      label: textOrUndefined(child(elem, 'label')),
      caption,
[...]

PR #674, opened 2020-09-08
fix(JATS): Decoding tag table-wrap-foot for table

nokome:
This is a copy of #667 created as a branch in this repo so we can all better collaborate on it.


nokome:
Hi @asisiuc. I have made some changes to make the table description an array of block content nodes. I think this will be more useful when decoding to HTML, or other formats.

I will merge this now. Please feel free to submit another PR, if this structure does not meet your needs.

PR #689, opened 2020-09-17
Always coerce to schema when importing content

nokome:
This PR ensures that after decoding, all content is coerced to ensure that it conforms to the schema. The `coerce` function is now applied in the `load` (from string) and `read` (from file) functions.

Most of the time decoded content will already conform to the schema (because our typings conform to the schema). However, there are some instances where content is loaded "as is" e.g. from JSON and YAML. In these cases, there is a possibility for non-conforming data to be in memory and thus cause exceptions when it is encoded using another codec. One example of this was the YAML metadata in RMarkdown.

This relies on https://github.com/stencila/schema/pull/217. There are some changes in this PR to changes in the schema in that PR.


alex-ketch (commented):
Looks good so far, I have one note and one question:
- Note: The Schema version dependency needs to be updated once the corresponding Schema PR is merged.
- Comment: Do we need to call `super.load()` or something from the individual Codecs? Right now it's not super clear when and where these methods get invoked.

nokome:
> Comment: Do we need to call super.load() or something from the individual Codecs? Right now it's not super clear when and where these methods get invoked.

Good question. No, we don't need to call `super.load()`. Individual codecs must implement `decode` and `encode` and these get called from `load` / `dump` and `read`/ `write` and ultimately from the top level functions such as `convert`. We should probably make `decode` and `encode` private methods - they should not get called directly, they should only get called via `load` etc. I think this would only affect tests.

Issue #725, opened 2020-10-22
MD: programming language is incorrectly decoded


alex-ketch:
Given the following RMD:

````rmd
# Test

```{python fig2}

# R code here

```
````

The programming language is captured as `"programmingLanguage": "{python",`

alex-ketch:
@nokome This seems to be happening in the `md` codec's handling of `decodeCodeblock`, not `xmd` side.

nokome:
But the `XmdCodec` should have "unbraced" it here: https://github.com/stencila/encoda/blob/231933e536b5295f7f2d724e1cb53126b0f762ed/src/codecs/xmd/index.ts#L173

Anyway, I'll create an issue test and see.

nokome:
This was related to #717 and has been fixed. See PR #726

PR #726, opened 2020-10-22
XMD: code chunk language

nokome:
At present this simply adds a test for #725. The test passes - the decoding of the provided snippet of XMarkdown is as expected and produces a `CodeChunk` with the  correct `programmingLanguage`.

@alex-ketch it seems that you may have been using the Markdown codec to decode the snippet rather than XMarkdown. When using an `.md` extension the expected decoding is a Code**Block**. The parsing of the `programmingLanguage` in this case is handled by `remark` and it treats the opening curly brace as part of the language. Happy to consider alternative behaviour for the `MdCodec` (e.g. strip curly braces from the Commonmark ["info string"](https://spec.commonmark.org/0.29/#info-string)


alex-ketch:
Hmmm, interesting. Something must be happening in the `sniff` functionality then @nokome.
The command I was using was `npx ts-node --files src/cli.ts convert ./article-2.rmd --to json ./out-json.json`, note that I wasn't passing a `--from` argument.

nokome:
@alex-ketch I have added another test to check the `match` function but am still not able to find an issue there. Nor am I able to when I run the CLI on the test fixture file added as part of this branch: 

```
> npx ts-node --files src/cli "convert" "src/__tests__/issues/725-xmd-code-chunk-language.rmd" "-" "--to" "json"

{
  "type": "Article",
  "content": [
    {
      "type": "CodeChunk",
      "label": "fig1",
      "programmingLanguage": "r",
      "text": "plot(1:100)"
    }
  ]
}
```

Could you please try reproducing the issue on this branch. The only thing that I can think of is that the test file you are using has Windows line endings and you are using a version of Encoda prior to 231933e536b5295f7f2d724e1cb53126b0f762ed being merged in.

alex-ketch:
@nokome yup, probably was due to a file I downloaded with Windows line endings. Sorry for the wild goose chase 😅

nokome:
No worries, :duck: :runner:  I'll merge this anyway and make sure the Hub has the most recent release

PR #836, opened 2021-03-04
Crop out RPNG symbol indicator when decoding images

alex-ketch:
This is the counterpart PR to https://github.com/stencila/thema/pull/304.
Note that the working assumption for this PR is that RPNG with `contentUrl === 'data:self'` always have an RPNG symbol.
If that's not the case down the road, we could make the cropping optional.


nokome (approved):
LGTM

PR #842, opened 2021-03-15
Improve LaTeX codec and add `noweb` codec

nokome:
This PR improves the `latex` codec and adds a `xnw` codec to support adding code to LaTeX.

### Improvements to `latex` codec

Based on some initial tests, these are some of the issues we've identified with the `decode` method of the `latex` codec,

- [x] Should be the default for `tex` files
- [x] Handle `bibliography` property of article
- [x] Add space after periods, between words, after commas (check for other situations where that might happen) (aaf761a87841f52afed02e01c8da546722aca0cd)
- [x] Add space before citation parentheses
- [x] "Content" property of citation is consuming `\citep`, that is, "e.g." should be rendered as regular text rather than as part of the citation (1ecac4fef1b4574a6723b40c590dfc11f43fed16)
- [x] Number equations (1d348b4a9c8c5bda8bc41fe58c5d538bc01011a0)
- [x] ~~Table caption width sometimes match table width, rather than page width (for some styles, maybe not for all)~~ (this is an issue with the Nature theme in Thema, not related to the `latex` codec - see https://github.com/stencila/thema/issues/312)
- [ ] Table caption in html element is missing, probably to do with automatic numbering in LaTeX
- [ ] Fix reference to equations
- [ ] Box environment not being rendered at all

For now, we are not too concerned about the correctness of the `encode` method (i.e. converting to LaTeX).

### Add a `nw` codec

`noweb` is syntax for literate programming allowing embedding executable code within LaTeX:

- https://www.cs.tufts.edu/~nr/noweb/
- https://github.com/nrnrnr/noweb
- https://www.cs.tufts.edu/~nr/noweb/johnson-lj.pdf

We'll add a `nw` codec that can

- [x] Decode and encode `CodeChunk`s using the `<<..name of chunk...>>=` syntax
- [x] Decode and encode inline `CodeExpressions` using `\expr{...}` and `\Sexpr{...}` (for Rnw compatability) syntax
- [x] Infer the language  based on the file name extension e.g. `rnw`, `pynw`.

### Other

- [x]  Research whether there are any other conventions / syntaxs for incorporating executable code within LaTeX.


nokome:
@fkmendes I've started working on the labeling of equations. It seems that Pandoc does not parse out `\\label{}`. See this example, Pandoc JSON:

```json
              "t": "Math",
              "c": [
                {
                  "t": "DisplayMath"
                },
                "\\text{E}[x] = \\frac{1.0}{1.0}\n  \\label{eq:eq1}"
              ]
```

from this LaTeX:

```latex
\begin{equation}
  \text{E}[x] = \frac{1.0}{1.0}
  \label{eq:eq1}
\end{equation}
```

Furthermore, there are Pandoc extensions like https://github.com/tomduck/pandoc-eqnos which suggest confirm the suspicion that Pandoc does not handle these natively.

So what to do? I'm leaning toward having Encoda parsing and removing the ` \label{eq:eq1}` from the Latex and instead populating the `label` property of the `MathBlock` node. That would provide consistency with `Table` and `Figure` nodes and means we could have one function for doing automatic numbering of `Table`, `Figure` and `MathBlock` nodes (if the `label` property is not populated already.

nokome:
With respect to the "Content property of citation is consuming \citep ..." item above it seems to me that that is occurring, for me, because I don't have the package `natbib` installed. My first attempt at installing that on Ubuntu failed,

```sh
> sudo apt install texlive
> tlmgr install natbib

/usr/bin/tlmgr: unexpected return value from verify_checksum: -5
```

@fkmendes Any tips? Do you have `natbib` installed? Am I barking up the wrong tree?

nokome:
Will merge this now: there are still some items in the above list that are unresolved but they are probably best addressed as separate issues.

PR #868, opened 2021-03-19
Citation modes

nokome:
This PR adds awareness of the `citationMode` property of `Cite` nodes to the "main" codecs `pandoc`, `html`, `jats` and `md` as well  as the `reshape` function.

It is important to be able to differentiate between parenthetical and narrative citation modes when both `decode`ing and `encode`ing. The difference is qualitative and can not be dealt with by themes or by simply differentiating between `Cite` and `CiteGroups`. Thus far we have "gotten away" (partially) with ignoring the difference because of `(` etc characters in the text but this doesn't work well when moving across formats.

This was a larger (and more time consuming!!!) PR than I was anticipating but I think it's worthwhile to try and get these things right to save us time down the road.


src/codecs/md/plugins/cite.ts
```diff
@@ -2,90 +2,137 @@
...
+            .split(/\s*;\s*/)
+            .slice(1)
+            .map((cite) => (/^(.*?)@([\w-]+)(.*?)$/.exec(cite) ?? []).slice(1)),
+        ]
+      }
+
+      // Handle potentially more than one Cite
+      const items = groups.map(([prefix, target, suffix]) => {
+        if (prefix !== undefined) prefix = prefix.trim()
+        if (suffix !== undefined) suffix = suffix.trim()
```
alex-ketch, line 74: Can you give me some context on the need to `trim` these please?
Reason I ask is because this causes issues when going to other formats.
`[e.g. @ref1; @ref2 and others]` to html results in:
Screenshot 2021-03-19 at 12 03 23

Do you think something like this would be better?
```suggestion
        if (prefix !== undefined) prefix = prefix.trim() + ' '
        if (suffix !== undefined) suffix = ' ' + suffix.trim()
```

alex-ketch (changes requested):
Thanks for undertaking this big task @nokome 🙏 
I've spotted a couple of potential bugs that I've called out below.

src/__tests__/issues/737-org-authors.ts
```diff
@@ -29,13 +29,12 @@ test('issue 737: references with an organisational author', async () => {
...
+\tauthor = {{Malaria Genomic Epidemiology Network} and Rockett, KA and Clarke, GM and Fitzpatrick, K and Hubbart, C and Jeffreys, AE and Rowlands, K and Craik, R and Jallow, M and Conway, DJ},
 \tjournal = {Nature Genetics},
+\tyear = {2014},
+\tpages = {1197--1204},
 \ttitle = {Reappraisal of known malaria resistance loci in a large multicenter study},
 \tvolume = {46},
-\tauthor = {{Malaria Genomic Epidemiology Network} and Rockett, KA and Clarke, GM and Fitzpatrick, K and Hubbart, C and Jeffreys, AE and Rowlands, K and Craik, R and Jallow, M and Conway, DJ},
-\tpages = {1197--1204},
-\tdate = {2014},
-\tyear = {2014},
```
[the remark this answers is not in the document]
nokome: This and the next comment are actually unrelated to this PR (I pushed branches 'out of order'). They are related to an update to Citation.js.

Nonetheless, well spotted. Yes this was intentional as Citation.js is now only outputting `year` in BibTeX presumably because of aforementioned duplication .

src/__tests__/issues/737-org-authors.ts
```diff
@@ -29,13 +29,12 @@ test('issue 737: references with an organisational author', async () => {
   const bib = await new BibCodec().dump(ref)
   expect(bib).toBe(
     `@article{bib7,
+\tauthor = {{Malaria Genomic Epidemiology Network} and Rockett, KA and Clarke, GM and Fitzpatrick, K and Hubbart, C and Jeffreys, AE and Rowlands, K and Craik, R and Jallow, M and Conway, DJ},
```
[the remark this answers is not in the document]
nokome: Yes, I agree it's strange having all those `and`s but that is indeed [how BibTeX separates multiple authors](https://tex.stackexchange.com/a/36397) (and not necessarily how it appears in HTML or any other format). Again this is just the output from Citation.js.

src/codecs/md/plugins/cite.ts
nokome: The reason I trimmed was because the spaces after a prefix and before a suffix are not really part of the prefix or suffix but are rather separators from the `target`. 

Pandoc does the same for prefixes, but for suffixes "trims" the preceding white space to a single space (not sure why):

```
> echo '[prefix @ref1    suffix  ]' | pandoc --from markdown --to json | jq .blocks[0].c[0].c[0]
[
  {
    "citationId": "ref1",
    "citationPrefix": [
      {
        "t": "Str",
        "c": "prefix"
      }
    ],
    "citationSuffix": [
      {
        "t": "Space"
      },
      {
        "t": "Str",
        "c": "suffix"
      }
    ],
    "citationMode": {
      "t": "NormalCitation"
    },
    "citationNoteNum": 1,
    "citationHash": 0
  }
]
```

Currently separators are "added" to HTML via CSS in themes (as we do with separators between say authors and year in a `Cite`).

In Markdown, we add single separator spaces in: https://github.com/stencila/encoda/blob/926e49821c3a028478e1a34b3954823be5addfd7/src/codecs/md/index.ts#L1373

I'd prefer to leave as is.

alex-ketch: Thanks for clarifying. Given that we insert the whitespace in the Markdown codec, I'd prefer to avoid whitespace insertion in CSS unless we have to—as in the case of citations where we change element ordering.
Reason being that this causes issues when copy-pasting text, and we have already received a bug report for as much from a user.

PR #874, opened 2021-03-23
fix: Fix mixed citation in JATS

### LINKED ISSUE #873: JATS: Plain text citation
Mixed citations (e.g. originally from bibitems in LaTeX) are not parsed when reading from JATS-XML.

I think the `mixed-content` should actually be `mixed-citation`: https://github.com/stencila/encoda/blob/master/src/codecs/jats/index.ts#L1165

Even nicer would probably be to have all elements of the `mixed-citation` as inline elements to keep e.g. parts in italics.

```xml
<mixed-citation>Norman de Plume <italic>The book of XML problems</italic>. XtraPress 2021.</mixed-citation>
```

Maybe this could be filed as `description` or `comment`?

https://schema.stenci.la/creativework

rgieseke:
Closes #873

### DIFF
--- a/src/codecs/jats/__file_snapshots__/plosone-0091296.jats.xml
+++ b/src/codecs/jats/__file_snapshots__/plosone-0091296.jats.xml
@@ -1054,10 +1054,14 @@
                 </element-citation>
             </ref>
             <ref id="pone-0091296-Pitcher1">
-                <element-citation/>
+                <element-citation>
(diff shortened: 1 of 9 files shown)


nokome:
Thank you @rgieseke !

I have updated the test snapshots and tests now pass. You will see in the snapshots, that thanks to your change we are now capturing title e.g.

```yaml
  - type: Article
    id: pone-0091296-Choat1
    authors: []
    title: >-
      Choat JH (2012) Spawning aggregations in reef fishes; ecological and
      evolutionary processes. In: Sadovy de Mitcheson Y, Colin PL, editors. Reef
      Fish Spawning Aggregations: Biology, Research and Management. Heidelberg:
      Springer. pp. 85–116.
```

However, this obviously still needs improvement - but I will pick up that topic in the related issue.

PR #875, opened 2021-03-23
fix: Add equation id and alttext for MathML in JATS

### LINKED ISSUE #872: Equation ID
When going e.g. from JATS with MathML to JSON i noted that an id like

```xml
<inline-formula id="eq1"><mml:math xmlns:mml="http://www.w3.org/1998/Math/MathML" alttext="1\times 1" display="inline"><mml:mrow><mml:mn>1</mml:mn><mml:mo>&#xD7;</mml:mo><mml:mn>1</mml:mn></mml:mrow></mml:math></inline-formula> 
```
doesn't get an id in the resulting output.

I've changed the following
```diff
diff --git a/src/codecs/jats/index.ts b/src/codecs/jats/index.ts
index ef6a6e8c..40503729 100644
--- a/src/codecs/jats/index.ts
+++ b/src/codecs/jats/index.ts
@@ -2125,11 +2125,13 @@ function decodeMath(
     const image = decodeGraphic(graphic, inline)
     return inline ? image : [stencila.paragraph({ content: image })]
   }
-
+  const equationId = attr(formula, 'id') ?? ''
+  
   // Wrapper is needed to dump the entire math element
   const text = xml.dump(elem('wrapper', mathml))
   return [
     (inline ? stencila.mathFragment : stencila.mathBlock)({
+      id: equationId,
       mathLanguage: 'mathml',
       text,
     }),
```
Should i make a PR for this?

Another question, could the `alttext` attribute with the original TeX or screen reader information be a separate attribute?
If yes, would that have to be added in the Math schema (https://schema.stenci.la/math) or meta field?

rgieseke:
Closes #872

### DIFF
--- a/src/codecs/jats/__file_snapshots__/elife-43154-v2.yaml
+++ b/src/codecs/jats/__file_snapshots__/elife-43154-v2.yaml
@@ -2044,6 +2044,7 @@ content:
           width="thinmathspace"/><mml:mn>0.24</mml:mn><mml:mo>,</mml:mo></mml:mrow></mml:mstyle></mml:math>
       - ' then by the law of total probability:'
   - type: MathBlock
+    id: equ1
     mathLanguage: mathml
     text: >-
       <mml:math id="m1"><mml:mstyle displaystyle="true"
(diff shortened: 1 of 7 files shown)


nokome:
Thanks for this @rgieseke 🙏🏽 . I did some refactoring to make the resulting node only has properties that are
defined (i.e. no empty string for id, or altText) and updated the test snapshots to reflect the change (`npx jest -u`).

Issue #876, opened 2021-03-24
Textual/TeX representations of Math Nodes


rgieseke:
Following up from #872 
>I like the idea of capturing the alttext and agree that meta.altText is the best place for this for now. However, in the longer term it should probably live in a specific property (meta is really just a temporary dumping ground for properties that we don't have specific properties for. Some options could be
>
>    https://schema.org/speakable : although that seems to be intended for a URL or a 'content-locator' rather than the text itself
>    https://schema.org/alternativeHeadline: although not really an alternative title
>    or maybe creating a new alternativeText property which could also be used on CodeExpression, ImageObject etc
>
>Happy to consider alternatives. If you would like to progress this further a PR to stencila/schema would be appreciated.

I think there are different representations to consider. In many workflows the MathML comes from TeX. This might be worth keeping around as the final rendering might happen with KaTeX/MathJax from the TeX (even though they might also use/create MathML).

LaTeXML includes alttext as an attribute with the original TeX like this:

```xml
<disp-formula id="S0.Ex1">
  <m:math xmlns:m="http://www.w3.org/1998/Math/MathML" alttext="a=b^{2}" display="block">
    <m:mrow>
      <m:mi>a</m:mi>
      <m:mo>=</m:mo>
      <m:msup>
        <m:mi>b</m:mi>
        <m:mn>2</m:mn>
      </m:msup>
    </m:mrow>
  </m:math>
</disp-formula>
```
Pandoc creates Jats-XML and includes a separate [tex-math](https://jats.nlm.nih.gov/publishing/tag-library/1.2/element/tex-math.html) element:
```xml
<disp-formula>
<alternatives>
<tex-math><![CDATA[a = b^2]]></tex-math>
<mml:math display="block" xmlns:mml="http://www.w3.org/1998/Math/MathML"><mml:mrow><mml:mi>a</mml:mi><mml:mo>=</mml:mo><mml:msup><mml:mi>b</mml:mi><mml:mn>2</mml:mn></mml:msup></mml:mrow></mml:math>
</alternatives></disp-formula>
```

In the JATS tag browser `alt-text` is described like this:

>Accessibility: The short <alt-text> can be used for special accessibility display or presentation on graphic-limited websites or devices, as an alternative to providing the full graphic. (For example, the <alt-text> element is typically read by screen readers, and <alt-text> may also be used to display a few words “behind” a figure or graphic for devices with limited graphics capacity.)
[...]

PR #877, opened 2021-03-24
fix: Allow MathML elements with `m:math` namespace in JATS

rgieseke:
As can be seen in #876 tools like Pandoc or LaTeXML produce MathML nodes with different namespaces.

LaTeXML's `m:math` is currently not read by `encoda`.

A simple fix is to also look for this namespace in the JATS decoder. I looked into doing the right thing and trying to read the namespace but this is probably a good enough solution if you want to include this.

According to the JATS tag browser `m` is common even though `mml` seems to be the official standard.

> MathML Namespacing: The vast majority of users of this Tag Set use the DTDs as their primary (or only) model, so this Tag Set is maintained to ensure that users of DTDs and DTD-based tools will work well. Because DTDs do not “play well” with namespaces, and most DTD-based tools are not namespace aware, the MathML namespace has been hardcoded to the prefix “mml” in the DTDs. This means that the element name, in the DTDs, really is <mml:math>. While this was the usual prefix when this Tag Set was originally established, the more frequently seen prefix is now “m”. Since XSD and RNG use “real” namespaces, this makes no difference to them, as to a namespace-aware processor a namespace prefix is only a placeholder. For DTD users, the element names in the MathML will need to be “mml” (for example, <mml:mfrac>), but tools can be set to provide that prefix and the essence of namespaces is that prefixes do not matter, so no harm is done.

https://jats.nlm.nih.gov/publishing/tag-library/1.2/element/mml-math.html

Happy to look into providing test cases or consider alternatives!

### DIFF
--- a/src/codecs/jats/index.ts
+++ b/src/codecs/jats/index.ts
@@ -2117,7 +2117,7 @@ function decodeMath(
   formula: xml.Element
 ): (stencila.Math | stencila.ImageObject | stencila.Paragraph)[] {
   const inline = formula.name === 'inline-formula'
-  const mathml = first(formula, 'mml:math') || first(formula, 'm:math')
+  const mathml = first(formula, 'mml:math') ?? first(formula, 'm:math')
 
   if (mathml === null) {
     const graphic = first(formula, ['graphic', 'inline-graphic'])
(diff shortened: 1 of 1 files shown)


nokome:
Thanks for this @rgieseke, great to be getting your feedback and contributions!

I am happy to merge without a test given this is small but could you please fix the linting error:

https://dev.azure.com/stencila/stencila/_build/results?buildId=10972&view=logs&j=ad07b744-5a1a-5d54-7693-fcb02e666633&t=94df1b64-f1bb-5c49-0094-5b7458b33f31&l=20

by changing `||` to `??`.

rgieseke:
Sure, and will try to run `npm run lint` before pushing next time!

PR #879, opened 2021-03-25
feat(JATS): Read ids of paragraphs

rgieseke:
Here's a proposal to read ids from Paragraphs in JATS XML.

I've auto-updated the fixtures as well.

Turns
```xml
  <body>
    <p id="p1">Text<fn id="idm6"><p id="footnote1">Footy note.</p></fn></p>
  </body>
```

into

```json
  "content": [
    {
      "type": "Paragraph",
      "id": "p1",
      "content": [
        "Text"
      ]
    },
    {
      "type": "Paragraph",
      "id": "footnote1",
      "content": [
        "Footy note."
      ]
    }
  ]
```

(Adding support for footnotes <fn> would probably be a separate useful thing, but for me this would already help identifying footnotes and be useful in general i believe when reading JATS.)

### DIFF
--- a/src/codecs/jats/__file_snapshots__/f1000-8-1394-v1.yaml
+++ b/src/codecs/jats/__file_snapshots__/f1000-8-1394-v1.yaml
@@ -1394,6 +1394,7 @@ content:
             conditions.
     description:
       - type: Paragraph
+        id: FN1
         content:
           - type: Superscript
             content:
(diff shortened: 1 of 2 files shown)


nokome:
Thanks @rgieseke . I made some small refactoring in d451212bcb6a079ea55bbc52ecf83b47f6048fe4 to make the change more idiomatic (w.r.t to most of the rest of the repo).

PR #881, opened 2021-03-25
Add Vega codec, and support for Vega figure output to HTML Codec

### LINKED ISSUE #841: Add Vega codec
Vega is ["a visualization grammar, a declarative language for creating, saving, and sharing interactive visualization designs"](https://vega.github.io/vega/) based on JSON with a [JSON Schema](https://github.com/vega/schema) :sparkling_heart: . There are a couple of ways of generating the declarative JSON (other than by hand) including the Python package `altair`.

There has been interest from users in using Altair with Stencila and, more generally, Vega-Lite is likely to be the basis of any plotting functions that we add "natively" to Stencila (e.g. via a resurrection of [`mini`](https://github.com/stencila/mini)).

According to the [Altair docs](https://iliatimofeev.github.io/altair-viz.github.io/user_guide/display_frontends.html#renderer-api), there are four media types that we would probably want to support. It may be that these could all be handled by a single `vega` codec. Alternatively, we may want to split them out into two (or four :scream:).

-Vega-Lite 2.x: application/vnd.vegalite.v2+json
-Vega-Lite 1.x: application/vnd.vegalite.v1+json
-Vega 3.x: application/vnd.vega.v3+json
-Vega 2.x: application/vnd.vega.v2+json

The proposed codec would be very similar to the existing [`plotly` codec](https://github.com/stencila/encoda/blob/master/src/codecs/plotly/index.ts) (see the discussion in the PR that introduced the Plotly codec: https://github.com/stencila/encoda/pull/746): it would add the Vega JSON as `data` in the `content` of an `ImageObject`. In addition, we would need to handle the above media types in the [`encodeImageObject` of the `html` codec](https://github.com/stencila/encoda/blob/2cbe99eeb79ede286d5b3deacb8b22650f6456f0/src/codecs/html/index.ts#L2501).

alex-ketch:
Close #841
Supports https://github.com/stencila/designa/pull/181

This PR is mostly there, but I've got some bits in progress around the Vega version detection based on the spec/schema that I'd like to finalize before merging.

### DIFF
--- a/src/__fixtures__/image/vega.ts
+++ b/src/__fixtures__/image/vega.ts
@@ -1,0 +1,34 @@
+import * as schema from '@stencila/schema'
+import { vegaMediaType } from '../../codecs/vega'
+
+export const testData = {
+  description: 'A simple bar chart with embedded spec.',
+  spec: {
+    values: [
+      { a: 'A', b: 28 },
+      { a: 'B', b: 55 },
+      { a: 'C', b: 43 },
+      { a: 'D', b: 91 },
+      { a: 'E', b: 81 },
+      { a: 'F', b: 53 },
+      { a: 'G', b: 19 },
+      { a: 'H', b: 87 },
+      { a: 'I', b: 52 },
+    ],
+  },
+  mark: 'bar',
+  encoding: {
+    x: { field: 'a', type: 'nominal', axis: { labelAngle: 0 } },
+    y: { field: 'b', type: 'quantitative' },
+  },
+}
+
+export const vegaImage = schema.imageObject({
+  contentUrl: 'https://via.placeholder.com/300x150',
+  content: [
+    {
+      mediaType: vegaMediaType,
+      spec: testData,
+    },
+  ],
+})
--- a/src/__fixtures__/image/vega.ts
+++ b/src/__fixtures__/image/vega.ts
@@ -3,7 +3,7 @@ import { vegaMediaType } from '../../codecs/vega'
 
 export const testData = {
   description: 'A simple bar chart with embedded spec.',
-  spec: {
+  data: {
     values: [
       { a: 'A', b: 28 },
       { a: 'B', b: 55 },
--- a/src/codecs/html/index.ts
+++ b/src/codecs/html/index.ts
@@ -43,6 +43,7 @@ import * as vfile from '../../util/vfile'
 import { plotlyMediaType } from '../plotly'
 import { TxtCodec } from '../txt'
 import { Codec, CommonEncodeOptions } from '../types'
+import { isVegaMediaType, vegaMediaType } from '../vega'
 
 export const stencilaItemType = 'data-itemtype'
 export const stencilaItemProp = 'data-itemprop'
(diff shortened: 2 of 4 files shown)


alex-ketch:
Would appreciate your input on how to handle the various ways of encoding Vega versions @nokome when you have the chance.

---

I think we have two primary use cases of the mime/media types:
1. Identifying that a certain `ImageObject` content is a Vega figure
2. Figuring out the library variant and version to load when attempting to render the figure

Here are the various permutations they can take:

``` ts
const vegaMediaTypes = [
  // Our custom generic identifier. Not used by Altair or Vega libraries as far I'm aware.
  'application/vnd.vega',

  // Altair MIME bundle types
  // Note that it doesn't support the latest versions of Vega/Vega Lite libraries
  // @see https://iliatimofeev.github.io/altair-viz.github.io/user_guide/display_frontends.html#renderer-api
  'application/vnd.vegalite.v2+json',
  'application/vnd.vegalite.v1+json',
  'application/vnd.vega.v3+json',
  'application/vnd.vega.v2+json',

  // Used in Vega object `spec.$schema`, and can have many permutations.
  // @see https://github.com/vega/schema
  'https://vega.github.io/schema/[library]/v[versionNumber].json',
]
```

The question I have is whether we should encode things using the generic mime type (first in the list above), and enrich the `spec` object with the full url to the schema version? This makes reading extracting the version number a touch more involved, but straightforward nevertheless.
And if the mode/version is missing, should we try to hardcode in the current latest vega version number for future reproducibility?

nokome:
> The question I have is whether we should encode things using the generic mime type (first in the list above), and enrich the spec object with the full url to the schema version?

Yes, that seems like the right approach to me. Although I think the generic should be 'application/vnd.vega+json'.

> And if the mode/version is missing, should we try to hardcode in the current latest vega version number for future reproducibility?

Ideally the `spec` should always have a `$schema` property with the URL of the Vega/Vega-Lite schema. So I'm not sure that we need to parse the MIME type to determine the mode or version. I can check the `spec` produced by `altair`. i.e. maybe we just need to match `application/vnd.vega` in the MIME type and let Vega embed do the rest.

alex-ketch:
I think the we do still need to figure out the library versions to load the correct library.
Vega-embed figures out the right calls to make, but it doesn't handle loading of the correct library versions (meaning I need to rework the Designa component too).

See https://github.com/vega/vega-embed#directly-in-the-browser

> ## Directly in the Browser
> 
> You can import Vega-Embed from a local copy or (as shown below) from jsDelivr. Please replace [VERSION] with the correct Vega, Vega-Lite, and Vega-Embed versions. We recommend that you specify the major versions (vega@5, vega-lite@4, vega-embed@6).
> 
>     <!DOCTYPE html>
>     <html>
>     <head>
>       
>       <script src="https://cdn.jsdelivr.net/npm/vega@[VERSION]"></script>
>       <script src="https://cdn.jsdelivr.net/npm/vega-lite@[VERSION]"></script>
>       
>       <script src="https://cdn.jsdelivr.net/npm/vega-embed@[VERSION]"></script>
>     </head>
>     <body>

nokome:
Ah OK, the I suggest that you parse the `$schema` property of the `spec` for mode and version, not the MIME type. From the docs I'm pretty sure that `$schema` should always be present in the `spec`.

nokome (approved):
LGTM, thanks!

PR #890, opened 2021-03-30
feat(JATS): Decode ids from `graphic` objects

rgieseke:
This adds ids from JATS XML (if available) to graphic objects.


nokome (approved):
Thanks, LGTM!

Issue #880, opened 2021-03-25
JATS: Footnote support


rgieseke:
Probably related to issue #144 

How would one add support to read a footnote as an inline element as part of a paragraph?
Right now it's rendered as another paragraph _next_ to the paragraph which has the annotated word as the last word.

E.g. JATS:
```xml
    <p id="p1">Text<fn id="idm6"><p id="footnote1">Footy note about Text.</p></fn> begins and ends.</p>
```
To JSON:
```json
"content": [
    {
      "type": "Paragraph",
      "id": "p1",
      "content": [
        "Text"
      ]
    },
    {
      "type": "Paragraph",
      "id": "footnote1",
      "content": [
        "Footy note about Text."
      ]
    },
    {
      "type": "Paragraph",
      "content": [
        " begins and ends."
      ]
    }
```
It should be inline like a mark i think:
```json
  {
      "type": "Paragraph",
      "id": "p2",
      "content": [
        "Normal Text, ",
        {
          "type": "Strong",
          "content": [
            "bold"
          ]
        },
        " or not."
      ]
    }

```
I played around with adding a `decodeFootnote` function but don't really know what the best approach would be.

nokome:
Thanks for raising this @rgieseke. I propose a new `Note` type that would encompass footnotes and endnotes (and perhaps default to being an inline note i.e. in HTML it appears when you hover over it?):

```json
[
  {
    "type": "Paragraph",
    "id": "p1",
    "content": ["Text"]
  },
  {
    "type": "Note",
    "id": "idm6",
    "noteType": "Footnote",
    "content": [
      {
        "type": "Paragraph",
        "id": "footnote1",
        "content": ["Footy note about Text."]
      }
    ]
  },
  {
    "type": "Paragraph",
    "content": [" begins and ends."]
  }
]
```

The `Note.schema.yaml` in https://github.com/stencila/schema/tree/master/schema would look something like,

```yaml
title: Note
'@id': stencila:Note
extends: Entity
role: secondary
status: unstable
description: .
properties:
  noteType:
    '@id': stencila:noteType
    description: Determines where the note content is displayed within document.
    type: string
    enum:
      - Footnote
      - Endnote
  content:
    '@id': stencila:content
    description: Content of the note, usually a paragraph.
    type: array
    items:
      $ref: BlockContent
    $comment: |
      Most notes will have a single paragraph but could have multiple
      paragraphs, tables and even figures.
required:
  - content
```

[...]

PR #896, opened 2021-04-06
feat(JATS): Add footnote

rgieseke:
A first attempt at adding footnote (`fn`) support to the JATS decoder (as discussed in #880 and https://github.com/stencila/schema/pull/251) .

Considering the following JATS example:
```xml
<?xml version="1.0"?>
<article>
<body>
<p>Further<fn id="note-1"><p id="Ch0.note-1">An <bold>important</bold> comment.</p></fn> thoughts below.</p>
</body>
</article>
```
I get this JSON:
```json
{
  "type": "Article",
  "content": [
    {
      "type": "Paragraph",
      "content": [
        "Further"
      ]
    },
    {
      "type": "Note",
      "id": "note-1",
      "content": [
        {
          "type": "Paragraph",
          "id": "Ch0.note-1",
          "content": [
            "An ",
            {
              "type": "Strong",
              "content": [
                "important"
              ]
            },
            " comment."
          ]
        }
      ]
    },
    {
      "type": "Paragraph",
      "content": [
        " thoughts below."
      ]
    }
  ]
}
 ```
 Ideally, the footnote would appear as an inline element. Is this due to how paragraphs are handled?

### CODE REVIEW

**Comment by nokome** [comment]
Thanks for this @rgieseke !

> Ideally, the footnote would appear as an inline element. Is this due to how paragraphs are handled?

I decided in Schema that the content of a `Note` should allow for `BlockContent` only and inline nodes could get wrapped into a paragraph as you have done. We could expand the schema for content to include inline nodes, so that the wrapping is not necessary. But when encoding to another format e.g HTML it makes the code a little more cumbersome. Generally, I'm leaning towards making schemas as narrow as possible, while still be useful - mostly because it requires less type checking code when it comes to implementations. Ref https://github.com/stencila/schema/blob/9217c1435f92844f91cade18efefbf75b1854297/schema/Note.schema.yaml#L27

Could you please add `noteType: 'Footnote'` (that should probably be the `default` in the schema).

**Comment by rgieseke** [comment]
What i was expecting is something like
```json
{
  "type": "Article",
  "content": [
    {
      "type": "Paragraph",
      "content": [
        "Further",
        {
          "type": "Note",
          "id": "note-1",
          "content": [
            {
              "type": "Paragraph",
              "id": "Ch0.note-1",
              "content": [
                "A footnote",
              ]
            }
          ]
        },
        " thinking about it."
      ]
    }
  ]
}
```

I though a footnote is like something in parens which can also be read inline. Is this not possible because a block element cannot contain further block elements? The note itself could be an inline element, couldn't it?

rgieseke:
> Could you please add noteType: 'Footnote' (that should probably be the default in the schema).

Done! I was reluctant in my first commit to add this but this probably makes sense.
In JATS endnotes seem to be encoded as `fn` within `fn-group` but could also be just table notes etc.
For the *article* usecase this default is probably perfectly fine.

nokome:
Oh sorry, I misunderstood your point about the note being in it's own paragraph. Now I see what you mean...

> The note itself could be an inline element, couldn't it?

Yes, that was what I envisioned too. I think this may have been caused by the fact that we forgot to add `Note` to the list of `InlineContent` nodes. I have just released a new version of Schema with that fixed: https://github.com/stencila/schema/commit/aa370b665f2fd43dd00e32b96838df33bce0d41c. Can you try with the new version please.

Also it would be good to add a test for this using the above JATS and expected JSON.

rgieseke:
Looks good!

I had to introduce a workaround related to https://github.com/stencila/schema/commit/ebc45600cccad68fbabbb83562ac72d4c660f64e

This also affects other decoders (HTML, Markdown, Gdoc, Ipynb) and i'm not sure if my change is the proper solution.

```json
{
  "type": "Article",
  "content": [
    {
      "type": "Paragraph",
      "content": [
        "Further",
        {
          "type": "Note",
          "id": "note-1",
          "noteType": "Footnote",
          "content": [
            {
              "type": "Paragraph",
              "id": "Ch0.note-1",
              "content": [
                "An ",
                {
                  "type": "Strong",
                  "content": [
                    "important"
                  ]
                },
                " comment."
              ]
            }
          ]
        },
        " thoughts below."
      ]
    }
  ]
}
```

rgieseke:
I tried adding a test file, but right now i can't get to run the tests (npx jest src/codecs/jats) without failure even if i revert to earlier schema versions.
Any advice on how to debug that?

nokome:
> Any advice on how to debug that?

Sorry for the late reply, been offline for a few days.

Given the fixes that need to be made due to stencila/schema@ebc4560, I think the most efficient approach is for me to merge this PR and then work on those fixes for JATS and the other codecs on `master`.

PR #904, opened 2021-04-12
Content typings and coercion

nokome:
Fixes related to more restrictive schema definition for the `content` property of `Article`, `Table`, `Figure` and `CodeChunk, the addition of more `MediaObject` types to `InlineContent`, and some other changes in schema.

Previously, an array of nodes `Node[]` was allowed as `Article.content` and `Table|Figure|CodeChunk.caption`. This was very permissive, allowing all sorts of `content` including plain JSON objects, arrays, and primitives. In https://github.com/stencila/schema/commit/ebc45600cccad68fbabbb83562ac72d4c660f64e and https://github.com/stencila/schema/commit/4acc3ba691a17844d7a24f1a295c3d352c83aac6 the schema for these was restricted to `BlockContent[]`. This was partly to be more consistent with recent additions e.g. [`Note`](https://github.com/stencila/schema/commit/b187519d72ed8d77fa85bc3c44abc1b81ad5d93a) and partly because when developing the Rust bindings in https://github.com/stencila/schema/pull/250, the `Node[]` typing caused problems (because it is so wide).

Narrowing the types for these properties is beneficial but had some unforeseen consequences on the coercion done by the `coerce` function. Previously, because the schema was so permissive, no coercion was done. Now, with a narrower schema coercion "kicks in" and strings like `"42"` get coerced to an `integer` etc. Again, this is intended and beneficial, but requires numerous updates to tests and test snapshots.


nokome:
Merging this now to get master green but with `[skip release]` because there are issues with Ajv's coercion rules that need to be addressed.

rgieseke:
Not sure whether it could become a problem but what about floats?
Something like `12345.6789` could also be an id as part of a link.
Might not be problematic but not sure how all used languages handle reading floats from JSON.

PR #909, opened 2021-04-14
feat(JATS): Add statement decoder

rgieseke:
Here is a first attempt at adding a JATS `statement` decoder.
It depends on adding `Claim` to the list of valid block elements in `schema`
https://github.com/stencila/schema/pull/261

The test case i've added likely requires some setup, i couldn't get it to run locally.

XML testcase:
```xml
<?xml version="1.0"?>
<article>
  <front>
    <article-meta>
      <contrib-group />
    </article-meta>
  </front>
  <body>
    <statement id="statement1">
      <label>Hypothesis 1</label>
      <p>It's a hypothesis.</p>
    </statement>
    <statement id="statement2">
      <title>Theorem 1</title>
      <p>It's a theorem.</p>
    </statement>
  </body>
  <back>
    <app-group />
  </back>
</article>
```
Encoda output:
```json
  "type": "Article",
  "identifiers": [],
  "title": "Untitled",
  "meta": {},
  "content": [
    {
      "type": "Claim",
      "id": "statement1",
      "claimType": "Statement",
      "label": "Hypothesis 1",
      "content": [
        {
          "type": "Paragraph",
          "content": [
            "It's a hypothesis."
          ]
        }
      ]
    },
    {
      "type": "Claim",
      "id": "statement2",
      "claimType": "Statement",
      "label": "Theorem 1",
      "content": [
        {
          "type": "Paragraph",
          "content": [
            "It's a theorem."
          ]
        }
      ]
    }
  ]
}                                     
```

I'm not 100% sure `label` and `title` should be treated as interchangeably as i did here, but given that the `title` usage seems to be the one in LaTeXML and `label` is the default example in the JATS [tag browser](https://jats.nlm.nih.gov/publishing/tag-library/1.2/element/statement.html) i think it's a reasonable start.

One could try parsing the statement `description` out of the `label` field, but maybe that's for later.


nokome:
Thanks @rgieseke.

I have created two tests snapshots by running `npx jest jats/ -u`. They look OK to me, but you might want to check them. Hopefully tests will all pass now.

It's tempting to parse the `claimType` out of the `label` or `title`, rather than using `Statement`. I might have a quick crack at that in the morning but will merge if it's not something that is easy to do.

rgieseke:
Thanks!

When running
`npx jest src/codecs/jats/jats.test.ts` or `npx jest jats` i get errors like

```
Property 'toMatchFile' does not exist on type 'JestMatchers<string>'.
```
Could be similar to this problem: https://github.com/jest-community/jest-extended/issues/189
(Though jest-extended is not used here.)

rgieseke:
> It's tempting to parse the claimType out of the label or title, rather than using Statement. I might have a quick crack at that in the morning but will merge if it's not something that is easy to do.

It would be nice indeed, only problem could be cases where special names are used (like "Fermat's Theorem" and not something like "Theorem 22").

nokome:
> i get errors like

Yes, me too. It's a pain. My workaround is to run `npx jest --clearCache`. But a longer term solution would very much nicer.

> Fermat's Theorem

Nothing a little regex couldn't handle ;) (not sure Fermat would approve though)

rgieseke:
> Yes, me too. It's a pain. My workaround is to run npx jest --clearCache. But a longer term solution would very much nicer.

:rocket: that does the trick!

> Nothing a little regex couldn't handle ;) (not sure Fermat would approve though)

:thinking: Maybe even an `includes` can get a long way ...

nokome:
I've added parsing of `claimType` from `label` and added more tests cases.

Thanks again for this ✨ , will merge now.

PR #912, opened 2021-04-16
fix(JATS): Read label or title as inline content

rgieseke:
Requires an update to InlineContent in `schema`. (https://github.com/stencila/schema/pull/262)
[skip ci]


rgieseke:
> I'd prefer to keep Claim.label a string, consistent with Figure and Table (there are advantages such as being able to generate and id from the label) , and instead look at alternatives for handling title in Encoda. I'll comment on the PR over there.
(https://github.com/stencila/schema/pull/262#pullrequestreview-639709185)

I think there are multiple options, the problem being that `label` and `title` appear to be used somewhat interchangeably "in the wild".
One option could be:
Read `label` and `title` as a string (and parse the `claimType`), but keep `title` as full nodes within content.

Another option:
Read `label` and `title` as a string (and parse the `claimType`) and have a separate `title` field (similar to a figure caption).
Could this be a `stencila.heading`? In JATS a `<title>` is
>Heading or title for a structural element (for example, <sec>, <app>, <boxed-text>).

nokome:
Sorry for the slow reply, been a tad busy with other things.

I agree with read `<label>` and `<title>` as a string and parse the `claimType`, which is what we already do:

```ts
const label = textOrUndefined(child(elem, ['label', 'title']))
...
```

But if there is anything other than just plain string content in the `<title>` (or `<label>`?)  set the `title` property of the `Claim` (it already has that property inherited from `CreativeWork` and it can contain inline content https://schema.stenci.la/Claim).

How does that sound?

rgieseke:
> Sorry for the slow reply, been a tad busy with other things.

No worries, that's what async discussions are for :-)

> But if there is anything other than just plain string content in the <title> (or <label>?) set the title property of the Claim (it already has that property inherited from CreativeWork and it can contain inline content https://schema.stenci.la/Claim).
> 
> How does that sound?

Sounds perfect! Is there already a way to detect that or just try an "decodeElements" and check whether it is an array?

nokome:
> Is there already a way to detect that or just try an "decodeElements" and check whether it is an array?

No special way I don't think, something like the following should suffice

```ts
const nodes = decodeElement(item, state)
let title = nodes.length == 1 && nodes[0] === 'string' ? nodes : undefined
```

rgieseke:
Here is an attempt to read the `title` as an inline element if it exists.
I would keep the `label` as is. Would get complicated to differentiate if the label exists and also a title.

nokome:
Thank you for the changes 👍🏽 , merging now.

PR #957, opened 2021-06-16
Fix releases: Ensure all project dependencies are contained inside src directory

alex-ketch:
ce34372095b0413c0cbf3043fab3b96feeb61cd3 broke the release as it caused TypeScript to output unexpected `dist` content structure.

- Last working version: [encoda@0.116.1](https://unpkg.com/browse/@stencila/encoda@0.116.1/dist/)
- Latest version (not working): [encoda@0.117.0](https://unpkg.com/browse/@stencila/encoda@0.117.0/dist/)

Root cause was due to the `codemeta.json` file living outside the `src` directory, resulting in `rootDir` value to resolve to `.`.

Manually setting `rootDir` to `./src` does not help because this results in the following error:
> 'rootDir' is expected to contain all source files.

The fix was to move the `codemeta.json` file into the `src` directory.


nokome:
Thanks for this @alex-ketch.

Unfortunately, if we move the `codemeta.json` file to `src/` then we won't be adhering to the [CodeMeta standard](https://codemeta.github.io/user-guide/) and thus loose any discoverability interoperability that may bring.

Instead, perhaps we need to do is have a `src/manifest.ts` file with a `Manifest` instance from which we generate the root `codemeta.json` as part of the build and commit for each release.

Note that if we change the `rootDir` back to `src` then we also need to remove the `src` from here:

https://github.com/stencila/encoda/blob/master/src/codecs/pandoc/binary.ts#L17-L39

alex-ketch:
I've updated the binary path per your comment just to see if that makes the tests pass.

> Unfortunately, if we move the codemeta.json file to src/ then we won't be adhering to the CodeMeta standard and thus loose any discoverability interoperability that may bring.

I don't have enough insight re. the CodeMeta project to say definitively, but I read the location of the file being at the root more of as an example than a standard.

I don't have enough context to make a decision on how to move forward, but if you could point me in the right direction I can take a stab :) Is the `manifest.ts` the way to go @nokome?

nokome:
This is done (elsewhere).

PR #968, opened 2021-07-06
feat(JATS): Read permissions element from figures

rgieseke:
Here is an attempt at reading the permissions/license of a figure in JATS analogous to the main article license.

I'm not entirely sure how a figure with multiple images and diverging licenses should be handled:

> Contents of Permissions: The sub-elements of <permissions> are repeatable. If multiple values are applicable (such as multiple copyright years), they should be tagged separately rather than as multiple values within a single element. Repeated values do not describe the source of the material, they describe the current Intellectual Property for the article or object. 

https://jats.nlm.nih.gov/publishing/tag-library/1.2/element/permissions.html

On https://jats4r.org/permissions there are examples with repeated `<permissions>` elements:

I would think however that it's better in such a case to attach the different licenses the individual `<graphic>` elements.

Handling multiple `<permissions>` could probably be added to `encoda` but handling only one license doesn't require much code changes :slightly_smiling_face: 
(I tried for a bit but not sure if it's worth adding.)

### DIFF
--- a/src/codecs/jats/__file_snapshots__/fig.jats.xml
+++ b/src/codecs/jats/__file_snapshots__/fig.jats.xml
@@ -77,6 +77,22 @@ http://dx.doi.org/10.1063/1.4762852.1</ext-link>
                 </caption>
                 <graphic xlink:href="" specific-use="print"/>
             </fig>
+            <p>Example 6</p>
+            <fig>
+                <label>Fig. 6</label>
+                <caption>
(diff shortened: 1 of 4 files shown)


nokome:
Thanks, @rgieseke this looks great. I'm going to merge as is and if needs be we can tackle the more complicated scenario of a figure with multiple images and diverging licenses some time later.

PR #969, opened 2021-07-06
fix(JATS): Don't duplicate footnote content in abstract

rgieseke:
Previously all `<p>`s in an abstract would be rendered. With footnotes
a paragraph can appear nested "inside" of a paragraph.

Before:
```json
{
  "type": "Article",
  "description": [
    {
      "type": "Paragraph",
      "content": [
        "The abstract can contain a footnote.",
        {
          "type": "Note",
          "id": "note-1",
          "noteType": "Footnote",
          "content": [
            {
              "type": "Paragraph",
              "id": "footnote1",
              "content": [
                "A footnote from the abstract."
              ]
            }
          ]
        }
      ]
    },
    {
      "type": "Paragraph",
      "id": "footnote1",
      "content": [
        "A footnote from the abstract."
      ]
    }
  ],
```

With using `children` instead of `all`:
```json
  "type": "Article",
  "description": [
    "The abstract can contain a footnote.",
    {
      "type": "Note",
      "id": "note-1",
      "noteType": "Footnote",
      "content": [
        {
          "type": "Paragraph",
          "id": "footnote1",
          "content": [
            "A footnote from the abstract."
          ]
        }
      ]
    }
  ],
```

I tried updating the fixtures, but couldn't get it to work, even with `--clearCache`.

### DIFF
--- a/src/codecs/jats/__fixtures__/footnote.xml
+++ b/src/codecs/jats/__fixtures__/footnote.xml
@@ -1,6 +1,11 @@
 <?xml version="1.0"?>
 <article>
+<front>
+<abstract>
+<p>The abstract can contain a footnote.<fn id="note-1"><p id="footnote1">A footnote from the abstract.</p></fn></p>
+</abstract>
+</front>
 <body>
-<p>Further<fn id="note-1"><p id="Ch0.note-1">An <bold>important</bold> footnote.</p></fn> thoughts below.</p>
+<p>Further<fn id="note-2"><p id="footnote2">An <bold>important</bold> footnote.</p></fn> thoughts below.</p>
 </body>
 </article>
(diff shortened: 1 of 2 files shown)


nokome:
Thanks for this fix. I have updated the snapshots locally and unfortunately it breaks the snapshot `src/codecs/plos/__file_snapshots__/pmed-1002858.yaml` by removing the abstract completely. I'll look into it some more - it may need on level of unwrapping of the abstract contents.

rgieseke:
That's ... a problem! I just tested adding more paragraphs to the abstact in `__fixtures__/footnote.xml` and those didn't disappear.

rgieseke:
I looked a bit into the plos problem, but couldn't figure it out. I thought the decoder should be able to handle an array of paragraphs as an abstract.

PR #972, opened 2021-07-07
feat(JATS): Add label field to MathBlocks

rgieseke:
This adds extracting the label field from a `disp-formula`.

https://jats.nlm.nih.gov/publishing/tag-library/1.2/element/label.html

Example:
```xml
<disp-formula id="S0.Ex1">
    <label>Special-Label</label>
    <mml:math xmlns:mml="http://www.w3.org/1998/Math/MathML" alttext="a=b" display="block">
        <mml:mrow>
            <mml:mi>a</mml:mi>
            <mml:mo>=</mml:mo>
            <mml:mi>b</mml:mi>
        </mml:mrow>
    </mml:math>
</disp-formula>
```

To 
```json
{
  "type": "MathBlock",
  "id": "S0.Ex9",
  "label": "Special-Label",
  "mathLanguage": "mathml",
  "text": "<mml:math xmlns:mml=\"http://www.w3.org/1998/Math/MathML\" alttext=\"a=b\" display=\"block\"><mml:mrow><mml:mi>a</mml:mi><mml:mo>=</mml:mo><mml:mi>b</mml:mi></mml:mrow></mml:math>",
  "meta": {
    "altText": "a=b"
  }
}
```

### DIFF
--- a/src/codecs/jats/__file_snapshots__/elife-52882-v2.yaml
+++ b/src/codecs/jats/__file_snapshots__/elife-52882-v2.yaml
@@ -5884,6 +5884,7 @@ content:
       - ' in the form:'
   - type: MathBlock
     id: equ1
+    label: (1)
     mathLanguage: mathml
     text: >-
       <mml:math
--- a/src/codecs/jats/__fixtures__/equation.jats.xml
+++ b/src/codecs/jats/__fixtures__/equation.jats.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0"?>
+<article>
+    <front>
+        <article-meta>
+            <contrib-group />
+        </article-meta>
+    </front>
+    <body>
+        <p id="p1">A formula which can be referenced and has a label.</p>
+        <p>
+            <disp-formula id="S0.Ex1">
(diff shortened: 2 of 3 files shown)


rgieseke:
I think the breaking pipeline is due to some unrelated changes in `unist` types, i couldn't really figure out what changed (the recent commits in https://www.npmjs.com/package/@types/unist didn't seem related to the apparent necessary changes in https://github.com/stencila/encoda/blob/master/src/codecs/md/stringifyHtml.ts). 

As for `label`, in general this seems to work well in my tests though i'm not sure whether there should be a `decodeLabel` function.

nokome:
Thanks for another PR! Looks good.

> I think the breaking pipeline is due to some unrelated changes in unist types

Sorry, this was my bad, I merged in updates to dev dependencies without noticing that that particular PR way broken. I'll rollback that change and then merge this.

rgieseke:
Great, let me know if i can help somehow! I tried figuring out which dependency exactly broke out but wasn't successful. (Also not super easy to figure out what changed in which version exactly for these '@types' releases ...)

PR #1031, opened 2022-11-08
Extract additional information from JATS

nokome:
- [x] Extraction of missing ‘backmatter’ content from bioRxiv-style JATS XML
   - [x] Acknowledgements
      The `<ack>` tags id decoded to a new section in the `content` of the article e.g.
      ```json
      {
        "type": "Heading",
        "depth": 1,
        "content": [
          "Acknowledgements"
        ]
      },
      {
        "type": "Paragraph",
        "content": [
          "We thank..."
        ]
      }
      ```
   - [x] Other backmatter sections
      Other `<sec>` tags in the backmatter are added to the `content` of the article after acknowledgments.

- [x] Extraction of corresponding author information from bioRxiv-style JATS XML

  Now, `xref`s to `corresp` elements are followed in order to capture corresponding author emails.

- [x] Extraction of author contribution information from bioRxiv-style JATS XML

  Covered by extraction of `<sec>`s from backmatter (?).

- [x] Extraction of ORCID IDs for authors from bioRxiv-style JATS XML
   
   Added as an author identifier (in a similar manner to how DOIs are added as an article identifier) e.g. 
   ```json
      "identifiers": [
        {
          "type": "PropertyValue",
          "propertyID": "https://registry.identifiers.org/registry/orcid",
          "value": "http://orcid.org/0000-0002-4179-6995"
        }
   ```


nokome:
Commit d3e0057 attempts to extract emails from `<corresp>` elements linked by `<xref>`s. Unfortunately in some JATS articles the `<corresp>` element has multiple `<email>` elements and the current approach is only taking the first (meaning all corresponding authors get the same address - which is clearly wrong!). This is perhaps most easily fixed by assuming that the `<email>` elements are in the same order as the `<xref>`s that refer to them.

nokome:
> This is perhaps most easily fixed by assuming that the `<email>` elements are in the same order as the `<xref>`s that refer to them.

I've tried an approach that assume this (i.e. assuming the appearance order) but that can not be relied upon (e.g. in `src/codecs/jats/__fixtures__/81445f02-6bf5-1014-892a-9f5909cc27c1` the emails are in reverse appearance order. Another approach based on string matching the surname to the email or preceding text might be necessary.

PR #1084, opened 2024-10-16
Remove unused codecs and dependencies

nokome:
Some remaining opportunities for removing dependencies:

- [x] if CLI is not being used. i.e only being used by calling a function then `@stencila/jesta`, `get-stdin` and other dependenices can be removed
- [x] replace `@stencila/logga` with use of `console.warn`
- [ ] can the the `yaml` codec be removed (perhaps being used in tests)?


nlisgo:
Thanks for this @nokome. Great to catch up with you today. I will bring this up with the rest of the team today.

nlisgo:
This looks great. We have been getting some value from the cli in development. But we know the cli is over-engineered when we only support the conversion of JATS to JSON now.

I've submitted a PR to introduce a lightweight cli option which will reduce the feedback loop for us when developing support for things that aren't currently handled in Encoda.

https://github.com/stencila/encoda/pull/1085

I will merge in the above once green. I've tested the script on master and this feature branch and it works and gives the same output.

nlisgo (approved):
This looks good to me. The yaml removal could be done as a followup?

nokome:
I have rebased this on to current `master` and tests pass so will merge now as this dramatically reduces dependency count and code bloat.

PR #1086, opened 2024-10-18
Gather more values from reference elements if publication-type=other

nlisgo:


src/codecs/jats/index.ts
```diff
@@ -1375,11 +1383,6 @@ export function decodeReference(
       // Not a chapter so title is <source>
       title = textOrUndefined(child(elem, 'source'))
```
fred-atherden, line 1384: Do I understand correctly that  if publication type is `'software', 'web', 'patent'` and has a title derived from `['article-title', 'chapter-title', 'data-title']`, then this line will overwrite that title (when present) with `source`? 

Ideally we would support similar logic fo `'software', 'web', 'patent'` reference types. In other words (for those reference types) if there is one of `['article-title', 'chapter-title', 'data-title']` and a `source`, then the former is decoded as the title and the `source` is used to decoded it as part of 'something' (I'm not too concerned about what  personally - book or periodical - open to your thoughts/preference). 

If there's not one of `['article-title', 'chapter-title', 'data-title']` then the source is the title.

nlisgo: Might be best to go over this logic first thing on our call. In my adjustment to the code there should be not alteration, whatsoever, to the logic to determine the title.

The only adjustment to the logic should be that we attempt to set the `isPartOf` if the publication type is other in the same way as if it is journal or preprint.

nlisgo: If I understand you correctly then we could treat anything that isn't `publication-type=book` or `publication-type=report` in the same way as journal, preprint or other with regards to the preparation of `isPartOf`?

fred-atherden: Yeah, I'd be happy with that approach. It wouldn't be semantically correct in the JSON since some of these won't be periodicals (e.g. a software reference for a GitHub repo), but I'm fine with that if you are.

And presumably the `meta.type` could be relied upon by client if there's a need to display the types differently.

src/codecs/jats/__file_snapshots__/06908fc3-73df-1014-bb56-a21daa237ef0.json
```diff
@@ -1595,6 +1595,10 @@
         "type": "Date",
         "value": "2015"
       },
+      "isPartOf": {
+        "type": "Periodical",
+        "name": "Nature"
+      },
```
nlisgo, line 1601: @fred-atherden I am happy with this adjustment with the logic. It would be great for you to scan over all/some of the changes to the fixture in the __file_snapshots__ folder as they represent the alteration to the output of the conversion because of the adjustments we have made to the logic.

I have removed the addition of `meta.type` to the decoded reference so we can reduce the differences in this review. As soon as this is merged in I will prepare that PR as a followup.

src/codecs/jats/jats.test.ts
```diff
@@ -188,7 +188,44 @@ test.each([
...
+          familyNames: ["Dubnau"],
+          givenNames: ["D"]
+        }
+      ],
+      datePublished: {
+        type: "Date",
+        value: "2019"
+      },
+      title: "YfmK is a Novel Nε-lysine Acetyltransferase that Directly Acetylates the Histone-like Protein HBsu in Bacillus Subtilis",
+      type: "Article"
```
nlisgo, line 228: @fred-atherden if there are additional test cases that you want to add and confirm that we get the output you expect, this is where you can add them.

src/codecs/jats/__file_snapshots__/06908fc3-73df-1014-bb56-a21daa237ef0.json
fred-atherden: Thanks for this @nlisgo, I've taken a look at the changes to the file snapshots and they're looking good. We're surfacing more information as desired. I think we're good to go ahead with this - thanks for your help!

src/codecs/jats/jats.test.ts
fred-atherden: Thanks - all good to proceed without any additions

fred-atherden (approved):

PR #1088, opened 2024-10-24
Add the reference published year as a string to meta to preserve any suffixes

will-byrne:
Fix for https://github.com/elifesciences/enhanced-preprints-issues/issues/815 

Adds the unprocessed year published string to the meta if it is present to preserve the suffix without requiring a schema change.


src/codecs/jats/index.ts
```diff
@@ -1493,7 +1493,16 @@ export function decodeReference(
   ]
   if (identifiers.length === 0) identifiers = undefined
 
-  const meta = typeof label === 'string' ? { label } : undefined
+  const metaEntries: Record<string, string> = {}
+  if (
+    datePublishedString !== undefined &&
+    (datePublishedString.length > 4 || datePublished === undefined)
+  ) {
+    metaEntries.yearPublished = datePublishedString
```
nlisgo, line 1501: My suggestion is that we just store string in `meta.yearPublished` if it is defined.

```javascript
if (datePublishedString !== undefined) {
  metaEntries.yearPublished = datePublishedString
}
```

This will simplify the logic in client too. The only thing tricky in client is to have a fallback id `meta.yearPublished` is not defined to use `datePublished` until we do a full reimport.

